PR #498, opened 2017-01-10
Dotify 3.0 bundles

joeha480:
Updates dotify bundles to 3.0 versions. Note that, due to OSGi version range requirements, other bundles using the dotify bundles has to be updated as well. Therefore I have "unpinned" these, but you may do with them as you wish.


bertfrees:
Moved to PR in pipeline-mod-braille: https://github.com/daisy/pipeline-mod-braille/pull/139.

rdeltour:
@bertfrees was this in `v1.10`?

bertfrees:
It is currently still on the develop branch but will be in v1.10.1 of mod-braille (v1.10.2 of assembly).

PR #514, opened 2017-08-18
Update epubcheck and epubcheck-adapter

josteinaj:
- Update to working version of org.daisy:epubcheck that uses a newer version of guava
- Change epubcheck-adapter so that it uses org.daisy:epubcheck instead of org.idpf:epubcheck


bertfrees:
Thanks. Actually I already had prepared some of this in https://github.com/daisy/pipeline-assembly/pull/132, but that's OK. The reason it was pending is because we weren't sure whether we really need the "org.daisy" version. I only did two very small changes. Better would be if we could continue using the official version. In that case romain would need to make a release of it.

bertfrees:
@josteinaj Does this PR block something else?

josteinaj:
@bertfrees I need a working version of epubcheck and epubcheck-adapter for the nordic migrator, and I was getting some errors about the guava version. I'm still having some trouble getting all tests to run in the nordic migrator though; I'm getting a java exception for some of the tests. I will create a regression test for it, and if it turns out to be an issue with epubcheck-adapter or epubcheck I'll create a regression test there also.

If we want to keep using the IDPF-version of epubcheck, which would probably be best, then we need to fix the guava version issue.

@rdeltour: do you know if there's plans to upgrade to guava 15 in epubcheck?

bertfrees:
@rdeltour bump

josteinaj:
I believe I still need this to be able to make a release of the nordic migrator for v1.10.x.

@rdeltour @bertfrees is this on the TODO list?

bertfrees:
@rdeltour Are there plans to upgrade to guava 15 in epubcheck?

rdeltour:
> Are there plans to upgrade to guava 15 in epubcheck?

Not that I know.
If we need it we can create an issue and ask if other contributors/users are OK.

But Guava is now around v24, should we even make a bigger jump?

bertfrees:
OK, v24 makes sense.

bertfrees:
@rdeltour Do you want to create a epubcheck issue?

rdeltour:
> @rdeltour Do you want to create a epubcheck issue?

Done in IDPF/epubcheck#832

bertfrees:
@josteinaj This should have been fixed by https://github.com/daisy/pipeline-scripts-utils/commit/bd16b9b690f9990c40ecb1f932cb3bcb50f1d133 and https://github.com/daisy/pipeline-assembly/commit/0445bdc36870c6bc4d86c2f51079fe7de7afee07. Let me know if I can close this issue.

bertfrees:
@josteinaj I think I can close this issue right?

josteinaj:
Yeah

PR #518, opened 2017-09-08
[website] fix build error: 'No such file or directory'

josteinaj:
Fixed a build error for the website Makefile.

PR #519, opened 2017-09-11
Fix webui packaging

josteinaj:
Versioning of the webui doesn't work since the git tags are in the [daisy/pipeline-webui](github.com/daisy/pipeline) repo. So this changes it so that the webui is versioned explicitly instead.

Corresponding commit in the webui repo: https://github.com/daisy/pipeline-webui/commit/9530f89d114fe9fc37ac05fbf09561e7c0e60ceb

Branch also contains: #518

### DIFF
--- a/website/Makefile
+++ b/website/Makefile
@@ -1,4 +1,5 @@
 JEKYLL_SRC_DIR := src
+SHELL := bash
 JEKYLL_SRC_FILES_CONTENT := $(shell find $(JEKYLL_SRC_DIR)/{_wiki,_wiki_gui,_wiki_webui} -type f -not -name '_Sidebar.md' -not -name '*.png' )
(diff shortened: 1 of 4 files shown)


josteinaj:
I'll probably still release from daisy/pipeline-webui, but I probably could release from the super repo as well. This is just so that it's possible to build the DEB through the super repo (and thus build the docker image that I'm working on).

bertfrees:
OK so it also has an effect on the version of the Debian package?

josteinaj:
Yeah, the version of the DEB, and any other packaging of the webui, was generated based on tags in daisy/pipeline-webui.

bertfrees:
@joeha480 If you could merge https://github.com/daisy/pipeline-webui/pull/109 into develop first it would be easier to update the webui in the super project (I would simply have to commit the develop branch).

josteinaj:
@bertfrees ok, done.

bertfrees:
Thanks. Updated super project.

PR #520, opened 2017-09-11
Dockerfile

josteinaj:
Added a Dockerfile that builds the current commit.

The image is currently 1,7 GB in size. I've tried to reduce the size of the image by
cloning, building, installing, and removing the cloned repo in a single docker `RUN` step.
However, this introduces a limitation: the Dockerfile retrieves the current commit (`git rev-parse HEAD`)
from the remote, which means that any local changes won't be included in the build.
A simpler alternative would be to `ADD` all the files in the current repository to the image,
which would allow you to build local changes; but that might increase the size of the image.
So if this is a problematic limitation, we could try `ADD`ing the repo into the image.

Branch also contains: #518 #519

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,59 @@
+FROM ubuntu:16.04
+
+LABEL maintainer="jostein@nlb.no"
+
+WORKDIR /root/
+
+# Install dependencies
+RUN sed -i.bak 's/main$/main universe/' /etc/apt/sources.list
+RUN apt-get update && apt-get install -y locales && apt-get clean
+RUN locale-gen en_US en_US.UTF-8
+
+# Install Java
+RUN apt-get update && apt-get install -y openjdk-8-jdk && apt-get clean
+
+# Install other dependencies
+RUN apt-get update && apt-get install -y software-properties-common build-essential wget zip unzip git maven gradle golang libxml2-utils pcregrep && apt-get clean
+RUN apt-get update && apt-get install -y ruby ruby-dev zlib1g-dev && apt-get clean
+RUN gem install jekyll sparql rdf-turtle rdf-rdfa commaparty nokogiri mustache github-markup coderay
+
+# Initialize git repository and store hash of current git commit as an environment variable
+RUN mkdir -p /tmp/pipeline/.git/objects
+ADD .git/refs /tmp/pipeline/.git/refs
+ADD .git/HEAD /tmp/pipeline/.git/HEAD
+ADD .git/config /tmp/pipeline/.git/config
+RUN git --git-dir=/tmp/pipeline/.git rev-parse HEAD | sed 's/^/export SOURCE_COMMIT=/' | tee -a .bashrc >> .profile
+RUN mkdir -p pipeline && cd pipeline && git init
+RUN cp /tmp/pipeline/.git/config pipeline/.git/
+RUN cd pipeline && git remote | grep '/' | while read remote; do git remote remove $remote ; done
+RUN rm /tmp/pipeline -rf
+
+# Fetch, build and install Pipeline 2
+# (and run build commands with bash)
+RUN echo '#!/bin/bash -e -x\n\
+    cd ~/pipeline \n\
+    git fetch --all \n\
+    . /root/.profile \n\
+    git checkout $SOURCE_COMMIT \n\
+    make dist-deb \n\
+    tail -f maven.log & \n\
+    make dist-webui-deb \n\
+    DEBIAN_FRONTEND=noninteractive dpkg -i *.deb \n\
+    apt-get update && apt-get -f install && apt-get clean \n\
+    service daisy-pipeline2 stop \n\
+    service daisy-pipeline2-webui stop \n\
+    cd ~ \n\
+    rm pipeline -rf\n'\
+    >> /tmp/run.sh && \
+    sh /tmp/run.sh && \
+    rm /tmp/run.sh
+# in case the webui service didn't shut down properly:
+RUN rm /run/daisy-pipeline2-webui/play.pid
+
+EXPOSE 8181
+EXPOSE 9000
+
+# By default, start the Pipeline 2 Engine and Web UI when the container is started
+CMD service daisy-pipeline2 start && \
+    service daisy-pipeline2-webui start && \
+    bash
--- a/website/Makefile
+++ b/website/Makefile
@@ -1,4 +1,5 @@
 JEKYLL_SRC_DIR := src
+SHELL := bash
 JEKYLL_SRC_FILES_CONTENT := $(shell find $(JEKYLL_SRC_DIR)/{_wiki,_wiki_gui,_wiki_webui} -type f -not -name '_Sidebar.md' -not -name '*.png' )
 JEKYLL_SRC_FILES_MUSTACHE := $(shell find $(JEKYLL_SRC_DIR)/_data/_spines/ -type f -name '*.yml')
 JEKYLL_SRC_FILES_OTHER := $(filter-out $(JEKYLL_SRC_FILES_CONTENT) $(JEKYLL_SRC_FILES_MUSTACHE),\
--- a/webui/README.md
+++ b/webui/README.md
@@ -10,14 +10,9 @@ This project provides a Web User Interface for the DAISY Pipeline 2, developed w
 
 ### 1. Prepare the release
 
-The Web UI is versioned based on git tags. If you're on the `v2.0.0` tag,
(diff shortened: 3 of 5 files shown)


josteinaj:
It seems like there's a daisy organization (or user?) on docker hub already. @bertfrees @rdeltour are you the owner of this? https://hub.docker.com/u/daisy/

bertfrees:
I'm not the owner.

bertfrees:
> A simpler alternative would be to ADD all the files in the current repository to the image,
which would allow you to build local changes; but that might increase the size of the image.

Why do you think that might increase the size?

bertfrees:
How would customizing work? People just fork the super project? And when they don't want to use the super project they should create their own Dockerfile FROM this one and add any custom modules and config files?

rdeltour:
> It seems like there's a daisy organization (or user?) on docker hub already. @bertfrees @rdeltour are you the owner of this?

I'm not either…

bertfrees:
Could we claim it?

rdeltour:
> Could we claim it?

I guess you can always try :-)

josteinaj:
> I prefer to build the package beforehand and then ADD it

Yes, that would indeed require some build step. Maybe it could be set up by letting Travis do the building instead of Docker Hub.

> Why do you think that might increase the size?

If you `ADD` the entire git repository then it will be stored as a layer in the docker image. By cloning and deleting the repository in the same build step (`RUN`), the repository itself (which would probably include a lot of target directories with JARs) should not be stored in the image.

> How would customizing work?

If using the super repo, any branch will be built separately, and any forks could do the same. So for nlbdev/pipeline I would get a image which includes NLBs modules.

As long as branches/forks depend on unreleased (SNAPSHOT) versions of the modules, I don't think it's very convenient to use dockers built-in `FROM` mechanism.

bertfrees:
> Maybe it could be set up by letting Travis do the building instead of Docker Hub.

Is it possible to then upload the image to Docker hub? That would be nice. I guess that is what I had in mind when I said "deploy to Docker hub" in https://github.com/daisy/pipeline-assembly/issues/139#issue-254698923.

> If you ADD the entire git repository then it will be stored as a layer in the docker image.

OK, didn't know that. So is it not possible to delete what you have added in a RUN step?

> As long as branches/forks depend on unreleased (SNAPSHOT) versions of the modules, I don't think it's very convenient to use dockers built-in `FROM` mechanism.

Well, if they don't use the super-project it implies they depend on a released version of the pipeline. In that case they can use the FROM mechanism, and whether their own modules are released or not doesn't matter.

josteinaj:
> Is it possible to then upload the image to Docker hub?

Yes. You can push to docker hub from the command line. The thing where docker hub triggers a build on every commit having a Dockerfile at its root is just a convenience (you can also trigger on a regular interval, or if another docker image changes). If you want, you don't even need to have a Dockerfile, you can create a container, make changes, commit changes as layers (i.e. intermediate images), then push the image you're happy with to docker hub. Its been a while since I tried it though (a tutorial long ago), and a Dockerfile is definitely the way to go whether you build from Travis or docker hub.

> is it not possible to delete what you have added in a RUN step?

That's right. You can delete it, but it's still going to be stored in a previous commit/layer/image.

> Well, if they don't use the super-project it implies they depend on a released version of the pipeline. In that case they can use the FROM mechanism, and whether their own modules are released or not doesn't matter.

Right.

bertfrees:
Christian has created this: https://hub.docker.com/r/daisyorg/

bertfrees:
OK I'm going to merge this now. There are some improvements possible, like for example I would ideally like to see the server and webui split up into separate images and moved to the respective assembly and webui repos. We could then still have the option to combine them in a single image in order to run in localfs mode. Also the pulling from github is not ideal. But this is all secondary.

@egli Can you paste a link here to that new Docker feature you mentioned that allows you to do stuff in multiple passes?

egli:
https://docs.docker.com/engine/userguide/eng-image/multistage-build/

bertfrees:
@josteinaj Interesting?

egli:
re the organization that I created at dockerhub (https://hub.docker.com/r/daisyorg/): I'm basically just squatting the "domain" if you want. I'm happy to hand it over to the DAISY devs, if you'd like to use it.

josteinaj:
Oh, cool, native support for multistage builds! That indeed greatly simplifies things.

Then we can avoid the github cloning step, as we don't care about image size in the build image.

If we want separate engine and webui images, then those should probably go in daisy/pipeline-assembly and daisy/pipeline-webui. We'd have to include the DEB files in the super repo image so that they could be pulled into the pipeline-assembly/pipeline-webui docker images though.

bertfrees:
> We'd have to include the DEB files in the super repo image so that they could be pulled into the pipeline-assembly/pipeline-webui docker images though.

This I don't quite understand.

josteinaj:
> This I don't quite understand.

I thought docker hub only could have one Dockerfile per repository, and I figured the webui/assembly repos would need to pull in the DEB from the super repo using the multistage build mechanism. But actually there's a column for "Dockerfile Location" on Docker hub, so maybe we can build multiple images from the super repo.

bertfrees:
Yes, if we could build multiple images that would be ideal. But I still don't get what you wanted to do. Would webui and assembly depend on the super repo? Because that would be wrong. The dependency has to be the other way around, and my idea was to build the webui and assembly images from within the super repo when there are snapshot dependencies.

josteinaj:
> Would webui and assembly depend on the super repo?

Yes that's what I was thinking, but I see how that might be wrong.

egli:
Right now I have a very simple image to create the pipeline deb:
``` Dockerfile
FROM maven:3.5-jdk-8

ADD . /usr/src/daisy-pipeline2
WORKDIR /usr/src/daisy-pipeline2

# build Pipeline 2
RUN mvn clean package -P deb
```
It sits in the pipeline-assembly project. The idea is to convert this into a multistage build where the deb from the first stage is used in the second stage to install the pipeline.

I had thought that the webui would be a separate image that is based on the pipeline image. This would be built in a similar fashion with a multistage build.

I'm a bit confused as to where these Dockerfiles have to be located.

Would it be possible to have a quick chat either on irc or via jitsi?

egli:
Here's an updated version of the Dockerfile that builds and installs the pipeline using a multistage build:
```Dockerfile
FROM maven:3.5-jdk-8 as builder

LABEL maintainer="christian.egli@sbs.ch"

# build Pipeline 2
ADD . /usr/src/daisy-pipeline2
WORKDIR /usr/src/daisy-pipeline2
RUN mvn clean package -P deb


FROM openjdk:8-jre

# install Pipeline 2
WORKDIR /root/
COPY --from=builder /usr/src/daisy-pipeline2/target/daisy-pipeline2_1.10.3_all.deb .
RUN DEBIAN_FRONTEND=noninteractive dpkg -i daisy-pipeline2_1.10.3_all.deb

EXPOSE 8181

CMD service daisy-pipeline2 start && bash
```

This builds fine and creates an image with a reasonable size 
``` console
eglic@wharton:~/src/pipeline-assembly$ docker images
REPOSITORY               TAG                 IMAGE ID            CREATED             SIZE
daisy/pipeline2          latest              a52dfba26c23        About an hour ago   646MB
```
However when running the image it always seems to exit immediately
``` console 
eglic@wharton:~/src/pipeline-assembly$ docker run daisy/pipeline2
/etc/init.d/daisy-pipeline2: 62: /etc/init.d/daisy-pipeline2: ps: not found
pipeline2: JAVA_HOME not set; results may vary
OpenJDK 64-Bit Server VM warning: ignoring option PermSize=16M; support was removed in 8.0
OpenJDK 64-Bit Server VM warning: ignoring option MaxPermSize=256M; support was removed in 8.0
eglic@wharton:~/src/pipeline-assembly$ docker ps
CONTAINER ID        IMAGE               COMMAND             CREATED             STATUS              PORTS               NAMES
eglic@wharton:~/src/pipeline-assembly$ 
```

I get the feeling that the pipeline shouldn't even be run as a service inside docker. Instead it should just be started directly. So maybe after all I shouldn't install the deb package but simply the zip file (without the init scripts).

josteinaj:
@egli you should enable interactive (`-i`) and tty (`-t`). I also like to add a `--rm` so that the container is deleted after it's stopped. Try this:

```
docker run --rm -it daisy/pipeline2
```

josteinaj:
Already done? Close?

bertfrees:
I don't think I ever merged this. But building the Docker image is now done via a Github workflow  (see commit f16b64f3a3cc365b198c05a63341af928bf04715). See yes I think you can close.

PR #522, opened 2017-09-12
Updates dotify and braille-utils components

joeha480:


bertfrees:
Thanks. I checked that all test still pass. Will merge shortly.

joeha480:
I added another commit that should also be included

bertfrees:
Merged in 960b2bc.

PR #523, opened 2017-09-28
make SKIP_RELEASE=true

josteinaj:
This is to avoid errors like this during development or on CI servers:

```
[INFO]
------------------------------------------------------------------------
[INFO] Building braille-css 1.13.1-SNAPSHOT
[INFO]
------------------------------------------------------------------------
[WARNING]
/tmp/pipeline2/libs/braille-css/src/main/java/org/daisy/braille/css/RuleVolume.java:
/tmp/pipeline2/libs/braille-css/src/main/java/org/daisy/braille/css/RuleVolume.java
uses unchecked or unsafe operations.
[WARNING]
/tmp/pipeline2/libs/braille-css/src/main/java/org/daisy/braille/css/RuleVolume.java:
Recompile with -Xlint:unchecked for details.
--> test -e
.maven-workspace/org/daisy/braille/braille-css/1.13.1-SNAPSHOT/braille-css-1.13.1-SNAPSHOT.jar
--> bash .make/gradle-release.sh libs/braille-utils/braille-utils.api
Release libs/braille-utils/braille-utils.api manually. Then, run `make
-s dist-zip-linux' again to continue.

action suspended...
Makefile:253: recipe for target '.group-eval' failed
make: *** [.group-eval] Error 100
```

### DIFF
--- a/.make/gradle-release.sh
+++ b/.make/gradle-release.sh
@@ -2,6 +2,10 @@
 set -e
 set -o pipefail
 
+if [ "$SKIP_RELEASE" = "true" ]; then
+    exit 0
+fi
+
 make_cmd="make"
 if [[ -n $MAKEFLAGS ]]; then
     make_cmd+=" "
--- a/.make/mvn-release.sh
+++ b/.make/mvn-release.sh
@@ -2,6 +2,10 @@
 set -e
 set -o pipefail
 
+if [ "$SKIP_RELEASE" = "true" ]; then
+    exit 0
+fi
+
 make_cmd="make"
 if [[ -n $MAKEFLAGS ]]; then
     make_cmd+=" "


bertfrees:
I don't know... This doesn't really fix anything, just works around a bug.

I know where this is coming from. It must have started happening after commit https://github.com/daisy/pipeline/commit/202314857e29c6fa8b831d2a571ad3bc07513491. The problem is that some subrepos are behind. One of them is braille-utils.api. The version in libs/braille-utils/braille-utils.api/gradle.properties is 3.0.1-SNAPSHOT, but the version in modules/braille/maven/bom/pom.xml is 3.0.1. The Makefile currently tries to be smart and infers from this that the braille-utils.api snapshot needs to be released as 3.0.1. This was done for a reason, but I now think it is not so great, so I'm gonna try to find a better solution.

For now I think updating the braille-utils.api subrepo to the latest is a better way to get rid of the problem.

bertfrees:
You don't get the error locally because braille-utils.api-3.0.1 is in your local Maven repo already.

bertfrees:
I now get this error too. I first ran `make clean` which removes your .maven-workspace, so it makes perfect sense that in that case you get the error. For some reason this never happened before though.

For now I'm going to fix it by updating all the subrepos that are "behind". I have an idea about how to fix this properly, I'll do that later.

bertfrees:
@josteinaj Is this still an issue?

josteinaj:
No, this seems not to be an issue anymore.

PR #525, opened 2017-12-18
Document Docker distribution

bertfrees:


bertfrees:
@egli You said you'd mention the docker-compose method first because it's the easiest method. I made some changes. I didn't move it first because I want to start with the most basic configuration possible and then build up to a more advanced configuration. I think that makes it easier to follow. Also I'm not even so sure that docker-compose is the easiest method in case you only want to run the server. However I did make it a bit clearer that docker-compose is the recommended method for complex configurations.

src/Get-Help/User-Guide/Installation/index.md
```diff
@@ -67,6 +67,105 @@ The ZIP can be installed on all Linux distros. It includes the desktop
 application, the server and the command line tool. To install, simply
 extract the contents to any destination you want.
 
+### Docker
+
+The Docker distribution is not available as a download on the
+website. It comes in the form of a Docker image that you can obtain
+via the Docker command line interface:
+
+~~~sh
```
egli: I thought the images are pulled automatically if you type `docker run image`, so I wouldn't even mention the docker pull commands.
This of course depends on who your audience is. I envision someone like Mischa that doesn't care that much about the technical details and just want the pipeline up and running

src/Get-Help/User-Guide/Installation/index.md
```diff
@@ -67,6 +67,105 @@ The ZIP can be installed on all Linux distros. It includes the desktop
...
+docker run --detach \
+           -p 8181:8181 \
+           daisyorg/pipeline-assembly
+~~~
+
+This will make the web service available on the address
+http://localhost:8181/ws. By default, authentication is enabled. The
+key and secret are "clientid" and
+"sekret". [Environment variables]({{site.baseurl}}/Get-Help/User-Guide/Pipeline-as-Service/#environment-variables)
+can be provided with one or more `-e` arguments.
```
egli: I would add an example with the `-e` arguments

src/Get-Help/User-Guide/Installation/index.md
```diff
@@ -67,6 +67,105 @@ The ZIP can be installed on all Linux distros. It includes the desktop
...
+           -p 9000:9000 \
+           -e DAISY_PIPELINE2_URL: http://pipeline:8181/ws \
+           daisyorg/pipeline-webui
+~~~
+
+For running more complex configurations like these,
+[Docker compose](https://docs.docker.com/compose/) is recommended. A
+complete example, that also persists the application data, is given
+below. Simply create a file called "docker-compose.yml" with the
+following content and run `docker-compose up`.
```
egli: Why not simply refer to the `docker-compose.yml` file in the repo?

src/Get-Help/User-Guide/Installation/index.md
bertfrees: I had an example first but then removed it because there are examples below.

src/Get-Help/User-Guide/Installation/index.md
bertfrees: Ah yes you're right.

src/Get-Help/User-Guide/Installation/index.md
bertfrees: Dunno, I could refer to that as well, but it's nice to have it in the documentation. But what is it doing in the webui repo anyway? It doesn't have a real function (except for being linked to from the documentation). I also made some small changes by the way: removed the `PIPELINE2_WS_LOCALFS` because it's the default and only sensible setting and added the data volume for the server.

egli: > But what is it doing in the pipeline-webui repo anyway?

Good question. Is there a better place?

bertfrees:
> Everything is set up for you and you just have to direct your browser at it.

Again this is assuming you want to use the web UI.

The documentation should be understandable for an end user although in the majority of the cases it's gonna be a devop who reads it. End users will be more likely to reach for the desktop application. For someone who just wants to get the web UI running and doesn't care about anything else, like configuration settings etc., the recommended way is to download and run the Windows installer or Debian package. Maybe I could mention the docker-compose method on http://daisy.github.io/pipeline/wiki/webui/Installation-and-configuation?

src/Get-Help/User-Guide/Installation/index.md
bertfrees: Just on the website maybe. Or in a separate repo or gist. Or maybe in the pipeline-webui repo is fine if is linked to from the website.

If the docker-compose file would build the webui from source it would have a real function in the pipeline-webui.

egli:
> End users will be more likely to reach for the desktop application

That sounds reasonable

> Maybe I could mention the docker-compose method also on http://daisy.github.io/pipeline/wiki/webui/Installation-and-configuation but this time just say: "download this file and run docker-compose"?

That sounds like a good compromise

PR #539, opened 2019-05-17
daisy/pipeline-tasks#125 - compliance to epub 3.2 specification

NPavie:
Changes made to prepare the pipeline for the new [ePub 3.2 specification](https://w3c.github.io/publ-epub-revision/epub32/spec/epub-spec.html).

- Modified the mediatype mapping for the new media types compliant with epub 3.2
- Modified the metadata generation from zedai metadata process to complies with epub 3.2 specification.
- The epub3 metadata merging process raises a warning on metadata that are now deprecated with epub 3.2 specification
- The bindings generation process raises a warning, as this mechanism is deprecated in epub 3.2 specification and should be replaced by the standard HTML object fallback
- Fixed the generation of an empty dc:title tag when an empty doctitle tag was found in dtbooks.
- Fixed the generation of empty entries in the toc document of an epub when data-generated section or empty titles are found.
- New html-outline xproc step to encapsulate html5-outliner.xsl (the xsl is also removed from catalog)
- New corresponding xprocstep tests
- The toc creation step has been updated to use the new px:html-outline step

### DIFF
--- a/modules/common/html-utils/src/main/resources/META-INF/catalog.xml
+++ b/modules/common/html-utils/src/main/resources/META-INF/catalog.xml
@@ -4,8 +4,6 @@
           uri="../xml/xproc/html-library.xpl"/>
      <uri name="http://www.daisy.org/pipeline/modules/html-utils/library.xsl"
           uri="../xml/xslt/uri-functions.xsl"/>
-     <uri name="http://www.daisy.org/pipeline/modules/html-utils/html5-outliner.xsl"
-          uri="../xml/xslt/html5-outliner.xsl"/>
      <uri name="http://www.daisy.org/pipeline/modules/html-utils/html5-upgrade.xsl"
           uri="../xml/xslt/html5-upgrade.xsl"/>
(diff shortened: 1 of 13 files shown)


bertfrees:
@NPavie Thanks! I cherry-picked this commit on https://github.com/daisy/pipeline-modules/compare/epub3.2 (and modified it slightly by changing some version numbers). I also made a few small additional changes, such as the `file-uri` option on the `px:html-outline` step which I don't like that much. I also added some more tests to cover all the changes:

- px:epub3-nav-create-toc
- px:zedai-to-opf-metadata
- px:epub3-pub-create-package-doc

I know there didn't exist any tests for these steps yet, but a catch-up needs to be done, and the way I approach this is by adding tests for a step whenever it is touched, tests that cover at least the change that was made, and preferably also some other aspects of the behavior.

bertfrees:
Merged in 377b48a54f78fca5de1390cc14b9b9b5dee35582

PR #543, opened 2019-08-06
ACE validation module

NPavie:
New ACE validation utility module

Added a new internal step to allow the use of an existing ACE installation to create an accessibility report.
(the directory of the user's ACE installation must exist in the path for the provider to work)

The step should be available soon as an optional check in the epub3-validator script.


modules/scripts-utils/ace-adapter/src/main/java/org/daisy/pipeline/epub/AceProvider.java
```diff
@@ -0,0 +1,166 @@
...
+				File epubFile = new File(epubURI);
+				
+				if(aceProgram == null) throw new Exception("(running pxi:ace) ACE was not found in the PATH of your system");
+				
+				// constructing command line
+				String path = epubFile.getCanonicalPath();
+				ArrayList<String> arguments = new ArrayList<String>();
+				arguments.add(aceProgram.getAbsolutePath());
+				// Output where the ace reports (report.html and report.json) and unzipped epub will be stored
+				File tempDir = Files.createTempDirectory("ace-").toFile();
```
bertfrees: I'd call `createTempDirectory` inside the `else` of the `if` condition below.

modules/scripts-utils/ace-adapter/src/main/java/org/daisy/pipeline/epub/AceProvider.java
```diff
@@ -0,0 +1,166 @@
...
+				tempWriter.close();
+				// Read the resulting temp document and write it to the output xml report
+				// (reading the html report does not work here : the ace report is not strictly valid, 
+				// some non closed link tags are blocking the load by saxon within this code
+				outputResultNode.write(runtime.getProcessor().newDocumentBuilder().build(tempFile));
+			} catch (Exception e) {
+				logger.error("Exception raised while checking the epub with ACE : " + e.getMessage());
+				e.printStackTrace();
+				throw new XProcException(step.getNode(), e);
+			} finally {
```
bertfrees: Why the `finally`?

modules/scripts-utils/ace-adapter/src/main/resources/xml/xproc/library.xpl
```diff
@@ -0,0 +1,98 @@
...
+    <p:declare-step type="px:ace" name="main">
+        <p:documentation xmlns="http://www.w3.org/1999/xhtml">
+            <h1 px:role="name">ACE report generation</h1>
+            <p px:role="desc">Generate an accessibility check report using the Accessibility Checker for EPUB tool.</p>
+        </p:documentation>
+        <!-- anyFileURI to the epub -->
+        <p:option name="epub" required="true">
+            <p:documentation xmlns="http://www.w3.org/1999/xhtml">
+                <h1 px:role="name">EPUB URI</h1>
+                <p px:role="desc">URI of the EPUB to control</p>
```
bertfrees: "control" is not a good translation of "contrôler" in this context. "check" is better.

modules/scripts-utils/ace-adapter/src/main/resources/xml/xproc/library.xpl
```diff
@@ -0,0 +1,98 @@
...
+            </p:documentation>
+        </p:option>
+        <!--  -->
+        <p:option name="temp-dir" required="false" select="''">
+            <p:documentation xmlns="http://www.w3.org/1999/xhtml">
+                <h1 px:role="name">Temporary directory URI</h1>
+                <p px:role="desc">Intermediate directory where the unzipped epub and html and json ACE reports will be stored</p>
```
bertfrees: Maybe mention that by default a temporary directory will be created.

modules/scripts-utils/ace-adapter/src/main/resources/xml/xproc/library.xpl
```diff
@@ -0,0 +1,98 @@
...
+            </p:documentation>
+        </p:option>
+
+        <p:option name="output-dir" required="true">
```
bertfrees: When we discussed how to return the HTML report, I thought the idea was to return a `c:result` with the result URI, like you did, but I didn't know you were going to add also a `output-dir` option. I guess it's fine, but there's not really a reason for not specifying the full output path instead of the directory I think, or is there? The option could also be dropped completely, and the report could be put inside the `temp-dir` (this is what I had in mind initially).

modules/scripts-utils/ace-adapter/src/test/java/XProcSpecTest.java
```diff
@@ -0,0 +1,17 @@
+import org.daisy.pipeline.junit.AbstractXSpecAndXProcSpecTest;
+
+import static org.daisy.pipeline.pax.exam.Options.mavenBundle;
```
bertfrees: Some imports that can be dropped.

modules/scripts-utils/ace-adapter/src/main/resources/xml/xproc/library.xpl
```diff
@@ -0,0 +1,98 @@
...
+                    <p:with-option name="href" select="/c:result/text()" />
+                </px:html-load>
+                <!-- test with html load -->
+            </p:when>
+            <p:otherwise>
+                <p:in-scope-names name="vars"/>
+                <p:template>
+                    <p:input port="template">
+                        <p:inline>
+                            <jhove xmlns="http://hul.harvard.edu/ois/xml/ns/jhove" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" name="epubcheck-adapter" release="x.x"
```
bertfrees: Where did you get the idea to use a jhove report. Is it was ACE uses?

bertfrees: Also I'm thinking maybe we should mention in the step documentation that if ACE is not found on the system we get this warning and that the step does not throw an error.

modules/scripts-utils/ace-adapter/pom.xml
```diff
@@ -0,0 +1,69 @@
...
+    <artifactId>modules-parent</artifactId>
+    <version>1.12.1-SNAPSHOT</version>
+    <relativePath>../../parent/</relativePath>
+  </parent>
+
+  <artifactId>ace-adapter</artifactId>
+  <version>1.0.0-SNAPSHOT</version>
+  <packaging>bundle</packaging>
+
+  <name>DAISY Pipeline 2 module :: ACE Adapter</name>
```
rdeltour: It's a detail, but the proper case is "Ace", not "ACE" 😊 (applies here, and in other places like user messages in the code below)

modules/scripts-utils/ace-adapter/src/main/resources/xml/xproc/library.xpl
```diff
@@ -0,0 +1,98 @@
...
+        </p:option>
+
+        <p:option name="output-dir" required="true">
+            <p:documentation xmlns="http://www.w3.org/1999/xhtml">
+                <h1 px:role="name">Output directory URI</h1>
+                <p px:role="desc">URI of the folder were the accessibility check html report will be stored</p>
+            </p:documentation>
+        </p:option>
+
+        <p:output port="result" sequence="true">
```
rdeltour: maybe people would be interested in the JSON report as well, especially in automated workflows (the HTML report is mostly for end users to consume, while the JSON is for machine processing).

modules/scripts-utils/ace-adapter/pom.xml
```diff
@@ -0,0 +1,69 @@
...
+    <dependency>
+      <groupId>org.daisy.pipeline.modules</groupId>
+      <artifactId>html-utils</artifactId>
+      <scope>runtime</scope>
+    </dependency>
+  </dependencies>
+  
+  <properties>
+    <expose-services>
+      org.daisy.pipeline.modules.impl.Module_ace_adapter,
```
rdeltour: I'm not sure where this service comes from, but it's been a while I haven't implemented a step. @bertfrees is it generated code based on the XProc?

modules/scripts-utils/pom.xml
```diff
@@ -30,6 +30,7 @@
     <module>epub3-ocf-utils</module>
     <module>epub3-pub-utils</module>
     <module>epubcheck-adapter</module>
+    <module>ace-adapter</module>
```
rdeltour: again a detail, but the modules are ordered alphabetically for easier manual lookup :-)

modules/scripts-utils/ace-adapter/pom.xml
bertfrees: It is generated code based on the catalog.xml. The modules are now declaring/registering themselves rather than being discovered automatically by the framework. It's a change that was done in the context of the OSGi removal (although it's still possible in theory to discover modules on the class path, but this approach opens some perspectives).

rdeltour: Ok, that's what I suspected, sorry to have missed that. Thanks for the clarification!

bertfrees:
The documentation of the scripts that will get the "check-accessibility" option (or whatever you will call it) should probably have a link to the Ace installation instructions. But that's something to worry about later.

modules/scripts-utils/ace-adapter/src/main/resources/xml/xproc/library.xpl
NPavie: Well my idea was to output the content of the html report on the result port of the exposed step, and the uri from the internal step.
I can directly output the URI of both report (html and json) from the exposed step within the temp directory, which could become the only optionnal argument.
I'll update that on the next commit.

bertfrees: Ah yes I see, it's just an internal step. In that case I think it's indeed better to just drop output-dir.

modules/scripts-utils/ace-adapter/src/main/resources/xml/xproc/library.xpl
NPavie: For the ace adapter, i started from the epubcheck-adapter module which use this report (i assumed it was use by the pipeline to alert the user)
(And i noticed i forgot to update the name)
I agree with you on the added noticed in the documentation.

modules/scripts-utils/ace-adapter/pom.xml
NPavie: Oups, i thought it was spelled as an acronym, i'll change that 👍

modules/scripts-utils/ace-adapter/src/main/resources/xml/xproc/library.xpl
NPavie: Following Bert's mention, i'll update the output to be the URIs of both html and json report (like a `p:store` step)

modules/scripts-utils/ace-adapter/src/main/resources/xml/xproc/library.xpl
NPavie: Saxon cannot load the ace html report directly (i have to use the html-load step for that in the exposed step to load and return the html) so the `c:result` was just a way to return the data (similarly as a `p:store` step like we discussed on slack).

I'm thinking on using a fileset to return both html and json URIs (see @rdeltour request) instead of the `c:result` in the internal step.

bertfrees: Yes I know. But I was confused by the documentation of the result port. You are already returning the HTML content, but the documentation is wrong, it says it is a `c:result` document. This should be moved to the `result` port of the `pxi:ace` declaration.

I would use separate outputs for the HTML and the JSON. A fileset could be used for this in theory, the caller could filter it based on the media-type (html vs. json). But we normally use filesets for files that form a whole, like a DAISY or EPUB publication. This is a bit of a different use case I think, because we are interested in the individual files. What if, hypothetically, the files wouldn't be distinguishable based on media-type? What if the HTML or JSON or both aren't in the fileset, but you expect them to be?

NPavie:
I updated the code to now output both html and json reports (the json is wrapped in a `c:data` with reserved xml characters translated as entities) in two output ports (html-report and json-report).
I also added the "lang" option for reports localization.

I now use the `BinaryFinder` and `CommandRunner` classes as requested. I must warn you that the unix /dev/null path is used by default for output by the `CommandRunner` and that windows does not like it, so i had to do a small bypass by using an empty lambda.

I clarified the doc in the library but i still have to make a README.

modules/scripts-utils/ace-adapter/src/main/java/org/daisy/pipeline/epub/AceProvider.java
```diff
@@ -0,0 +1,181 @@
...
+				htmlReportURIport.write(runtime.getProcessor().newDocumentBuilder().build(tempFile));
+				tempFile.delete();
+
+				String jsonContent = new String(Files.readAllBytes(Paths.get(jsonReport.getAbsolutePath())),StandardCharsets.UTF_8)
+						.replaceAll("<", "&#x3e;")
+						.replaceAll(">", "&#x3c;")
+						.replaceAll("&", "&#x26;");
+				
+				if (!jsonContent.isBlank()) {
+					tempFile = File.createTempFile("json-data",".xml");
```
bertfrees: I would put this code for converting the JSON file to a `c:data` XML document in a separate function. It could even be made into a utility function in `file-utils` for instance, and it doesn't need to be specific to JSON (you could pass the content-type as argument). There could also be an XProc variant of this Java function, based on the following step internally declared inside `px:fileset-load`:

```xml
<p:declare-step type="pxi:load-text">
    <p:output port="result"/>
    <p:option name="href"/>
    <p:identity>
      <p:input port="source">
        <p:inline>
          <c:request method="GET" override-content-type="text/plain; charset=utf-8"/>
        </p:inline>
      </p:input>
    </p:identity>
    <p:add-attribute match="c:request" attribute-name="href">
      <p:with-option name="attribute-value" select="$href"/>
    </p:add-attribute>
    <p:http-request/>
  </p:declare-step>
```

Or it could use the Java implementation of course.

Maybe the code above for creating the `c:result` document could even be made into a common utility function. This is being done in several other steps so this is a way to simplify all that code.

I can do this myself.

modules/scripts-utils/ace-adapter/src/main/resources/xml/xproc/library.xpl
```diff
@@ -0,0 +1,162 @@
...
+                    </p:input>
+                </p:identity>
+                <px:html-load name="loading-html-report">
+                    <p:with-option name="href" select="/c:result/text()" />
+                </px:html-load>
+                <p:sink />
+            </p:when>
+            <p:otherwise>
+                <p:output port="html-report" primary="true"/>
+                <p:output port="json-report" />
```
bertfrees: Doesn't this cause an "unconnected port" error?

modules/scripts-utils/ace-adapter/src/main/java/org/daisy/pipeline/epub/AceProvider.java
```diff
@@ -0,0 +1,181 @@
...
+	 * Initialize the provider :
+	 * Retrieve the Ace executable path
+	 * @throws RuntimeException if Ace is not found on the system
+	 */
+	@Activate
+	public void init() {
+		Optional<String> lpath = BinaryFinder.find("ace");
+		if(lpath.isPresent()) {
+			aceProgram = new File(lpath.get());
+		} else throw new RuntimeException("ACE was not found on your system");
```
bertfrees: I'm not sure whether the exception should be thrown here or below in the `run` method. But probably not both.

bertfrees:
Great. The /dev/null thing, could that be fixed in the CommandRunner class? The lambda solution works, but it's better to fix the problem at the root. What is the proper way to do this on Windows?

The whole discussion about how to best return the HTML document, and the newly added JSON output, made me think about what is the best approach in general to return files. With a normal output port, with an option to pass the target file, an option to pass the target directory, an output port with a `c:result` document, an output port with a `d:fileset` document, etc. My conclusion is that it depends on the case, and sometimes there are multiple solutions. I would say that in this case, the more efficient way to do it would have been to return the file URIs of both HTML and JSON documents, because they are already stored to disk by Ace, and it is unlikely that the caller needs to modify the files. This is different from e.g. the px:epubcheck step which returns the report in an XML format. Of course you can also argue that the XProc language was designed to pass documents around in memory, and that the fact that Ace stores its outputs on disk is just an implementation detail. So the current solution is also fine.

@rdeltour Your view on this?

rdeltour:
> I would say that in this case, the more efficient way to do it would have been to return the file URIs of both HTML and JSON documents, because they are already stored to disk by Ace, and it is unlikely that the caller needs to modify the files

Actually the reports are stored on disk by Ace only when the `-o` option is specified. Otherwise, the HTML report isn't generated and the JSON report is written to `stdout`.

I think what we want for the XProc step largely depends on the targeted use case. If we don't know or want to keep it as open as possible, then we could produce the JSON in a `c:result` on the default output port, and control with options whether to produce the HTML report, and whether to send it to an output port in addition to storing it on disk.

bertfrees:
> control with options [...] whether to send it to an output port in addition to storing it on disk

Such an option I find unnecessary. It clutters/complicates the step interface while the loading (or storing) can easily be done by the caller.

To me it would even make more sense to split up the outputs into a fileset and a in-memory port (one for each output, html and json), and whether the files are stored in memory or on disk (at `href` or `original-href`) is then irrelevant, like should be the case for all our steps that have "fileset" and "in-memory" ports. The `px:fileset-load` and `px:fileset-store` steps worry about that.

But anyway since we are talking about single documents, going the fileset way is overkill really. Just deciding for each output whether it is stored on disk or in memory, and letting the caller load or store it if needed, is simpler in this case.

NPavie:
> Great. The /dev/null thing, could that be fixed in the CommandRunner class? The lambda solution works, but it's better to fix the problem at the root. What is the proper way to do this on Windows?

@bertfrees : On windows, the corresponding file is just 'nul', so quick fix would be to probe os type and swicth to file accordingly.
Another possibility is to assign consumers that does nothing (like i did in the AceStep) if they are undefined in the run method of CommandRunner.
That would pretty much does the same as sending the data to the void but without using OS-specific files.

For the output of the exposed step, if we plan on using the step to re-process an epub and automatically correct some accessibility issues, maybe the json would be more appriopriate indeed.
But i'm not sure how to process the json within xproc.
That's why i was loading the html report, even if it may be more complicated to use than the json one.

For the output of the exposed step, in that case, what i would propose is a single output port "result" port : 
it could output the json content as either `c:result` or `c:data` (i find `c:data` more approriate as encoding and content-type is defined, but if prefered i can use `c:result`)
A "reports-directory" option would specify where to put the ace reports if not used for processing.
For the internal step, eventually for future use, i could output the json result and the 2 reports URIs in 3 different port.

bertfrees:
If we want to use the accessibility support to do certain actions, it's maybe easiest to convert the JSON to some XML structure. By the way XProc 3.0 supports JSON natively (represented as XDM maps and arrays).

If you want to output the JSON content itself, `c:data` is indeed the way to go. But you can of course also output the file path when it's stored on disk. In that case you would use a `c:result`. However if the main use case for getting the JSON report is to interpret it in subsequent "fixer" steps, it is indeed more appropriate to output the content.

> A "reports-directory" option would specify where to put the ace reports if not used for processing. For the internal step, eventually for future use, i could output the json result and the 2 reports URIs in 3 different port.

As I said in the previous comment I find that this unnecessarily complicates things. Having two outputs for the same JSON document is just confusing. Also, steps such as `px:ace` shouldn't have to worry about the final location on disk of their outputs. That's the responsibility of "store"  and "copy" steps. So I wouldn't add a "reports-directory" option. I know that currently in other steps this approach is often used, but I think it's not the right approach.

NPavie:
> I find that this unnecessarily complicates things. Having two outputs for the same JSON document is just confusing. Also, steps like `px:ace` shouldn't have to worry about the final location on disk of their outputs. That's the responsibility of "store" and "copy" steps. So I wouldn't add a "reports-directory" option. I know that currently in other steps this approach is often used, but I think it's not the right approach.

So if i get it right : 
-  for the results of `px:ace`, should i go with the json report in a `c:data` as primary output, and the html report as a document in the secondary port ? 
- For the results of `pxi:ace`, should i go with the resulting html and json reports URIs as `c:result` ? 
(So i would load the content from the xproc step)

bertfrees:
What matters most for me is that the public step (`px:ace`) has two outputs, one for HTML and one for JSON. Whether these are the documents themselves, or the location on disk, is not so important. IMO it would have been simpler to just return the location on disk for the HTML output, but now that the `px:html-load` is already added there's no reason to change it back I think. I don't have a strong opinion about which output should be on the primary port. The JSON one is fine. Whether the content of the JSON document is loaded in Java or XProc also does not matter for me. Since it's already implemented in Java, I see no reason to change it now.

bertfrees:
OK I'm almost happy :-). But I can take over from here. I don't like putting the jhove document on the json port. I think you even got confused yourself because in the documentation you say that it is put on the html port. In fact I think we don't need jhove at all. I think what I will do instead is output some custom JSON like this:

```json
{ "error": "blah blah" }
```

for in case we want to catch this exception in the code and act upon it. An XML output would be slightly more convenient but I don't like allowing both an XML document and a non-XML document (which is what the `c:data` document should be seen as) on the same port. If we are going to allow JSON flow around in our pipelines anyway, we should be prepared to having to write utility steps for it (like converting JSON to a XML representation for instance).

On the html port we put a HTML document that says Ace was not found because the user needs to know this. The jhove report isn't really suitable for that because it isn't displayed directly in the web user interface. So we could convert the jhove to HTML, but then it's of course easier to generate the HTML directly within the `px:ace` step.

bertfrees:
@NPavie I've rebased the branch and pushed a couple of new commits to it. Please have a look and check if there is something you don't like. And maybe also quickly test if I fixed the "/dev/null" issue properly.

bertfrees:
Merged in 498680659a67f474378b66b38e1f14e19d5f655f

PR #545, opened 2019-10-09
html5-outliner

josteinaj:
Some minor improvements to the html5-outliner.

### DIFF
--- a/modules/common/html-utils/src/main/resources/xml/xslt/html5-outliner.xsl
+++ b/modules/common/html-utils/src/main/resources/xml/xslt/html5-outliner.xsl
@@ -47,9 +47,9 @@
     <!--  -->
     <xsl:template match="body|article|aside|nav|section">
         <xsl:param name="relative-path" tunnel="yes" as="xs:string" required="yes"/>
-        <xsl:variable name="id" select="@id"/>
         <xsl:variable name="heading" select="(h1|h2|h3|h4|h5|h6|hgroup)[1]" as="element()?"/>
         <xsl:variable name="heading-content" select="f:heading-content($heading,.)" as="item()*"/>
+        <xsl:variable name="id" select="(@id, $heading/@id)[1]"/>
         <xsl:variable name="children-doc">
             <xsl:copy-of select="* except $heading"/>
         </xsl:variable>
@@ -67,7 +67,8 @@
                     <a href="{$relative-path}#{$id}">
                         <!-- TODO: try to not "depend" on the TTS namespace here -->
                         <xsl:copy-of select="$heading/ancestor-or-self::*/@tts:*"/>
-                        <xsl:copy-of select="$heading-content"/>
+                        <xsl:copy-of select="$heading-content[count(.|../@*)=count(../@*)]"/>
+                        <xsl:copy-of select="$heading-content[self::* | self::text()]"/>
                     </a>
                 </li>
             </xsl:when>
@@ -89,7 +90,8 @@
                         <a href="{$relative-path}#{$id}">
                             <!-- TODO: try to not "depend" on the TTS namespace here -->
                             <xsl:copy-of select="$heading/ancestor-or-self::*/@tts:*"/>
-                            <xsl:copy-of select="$heading-content"/>
+                            <xsl:copy-of select="$heading-content[count(.|../@*)=count(../@*)]"/>
+                            <xsl:copy-of select="$heading-content[self::* | self::text()]"/>
                         </a>
                         <ol>
                             <xsl:call-template name="subsections">
@@ -212,7 +214,7 @@
     </xsl:function>
     
     <!-- -->
-    <xsl:function name="f:heading-content" as="item()*">
+    <xsl:function name="f:heading-content" as="node()*">
         <xsl:param name="node" as="element()?"/>
         <xsl:param name="parent" as="element()?"/>
         <xsl:choose>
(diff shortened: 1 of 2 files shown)


modules/common/html-utils/src/main/resources/xml/xslt/html5-outliner.xsl
```diff
@@ -67,7 +67,8 @@
                     <a href="{$relative-path}#{$id}">
                         <!-- TODO: try to not "depend" on the TTS namespace here -->
                         <xsl:copy-of select="$heading/ancestor-or-self::*/@tts:*"/>
-                        <xsl:copy-of select="$heading-content"/>
+                        <xsl:copy-of select="$heading-content[count(.|../@*)=count(../@*)]"/>
+                        <xsl:copy-of select="$heading-content[self::* | self::text()]"/>
```
bertfrees: Hmm I fail to see what this does. Can you explain?

modules/common/html-utils/src/main/resources/xml/xslt/html5-outliner.xsl
```diff
@@ -47,9 +47,9 @@
     <!--  -->
     <xsl:template match="body|article|aside|nav|section">
         <xsl:param name="relative-path" tunnel="yes" as="xs:string" required="yes"/>
-        <xsl:variable name="id" select="@id"/>
         <xsl:variable name="heading" select="(h1|h2|h3|h4|h5|h6|hgroup)[1]" as="element()?"/>
         <xsl:variable name="heading-content" select="f:heading-content($heading,.)" as="item()*"/>
+        <xsl:variable name="id" select="(@id, $heading/@id)[1]"/>
```
bertfrees: In a commit that is not on master yet I added the following documentation to `px:html-outline`:

> All <code>body</code>, <code>article</code>, <code>aside</code>, <code>nav</code>, <code>section</code>, <code>h1</code>, <code>h2</code>, <code>h3</code>, <code>h4</code>, <code>h5</code>, <code>h6</code> and <code>hgroup</code> elements must have an <code>id</code> attribute (see also px:html-id-fixer).

With your change, this would have to be changed to:

> All <code>h1</code>, <code>h2</code>, <code>h3</code>, <code>h4</code>, <code>h5</code>, <code>h6</code> and <code>hgroup</code> elements...

however `px:html-id-fixer` already takes care of `body` etc. too. Do you suggest we also change `px:html-id-fixer`, or we just leave it like it is?

modules/common/html-utils/src/main/resources/xml/xslt/html5-outliner.xsl
josteinaj: Good question. I'm guessing the intent is to exclude some types of content. This part can probably be reverted.

modules/common/html-utils/src/main/resources/xml/xslt/html5-outliner.xsl
```diff
@@ -225,19 +227,19 @@
                 <xsl:sequence select="f:heading-content($node/*[f:rank(.)=$rank][1],$parent)"/>
             </xsl:when>
             <xsl:when test="$parent[self::body]">
-                <xsl:sequence select="'Untitled document'"/>
```
josteinaj: Not really sure why I changed this either. Seems unnecessary.

modules/common/html-utils/src/main/resources/xml/xslt/html5-outliner.xsl
josteinaj: In case the books use ids on headlines instead of the surrounding sectioning element, a html outline can be created without modifying the document. Unless we want to do that, this is probably not necessary.

bertfrees: Well, I guess anything that make the conversion more robust is good. I think I'll pick this change and adapt the documentation. But I see no reason to change `px:html-id-fixer`.

modules/common/html-utils/src/main/resources/xml/xslt/html5-outliner.xsl
bertfrees: OK.

modules/common/html-utils/src/main/resources/xml/xslt/html5-outliner.xsl
bertfrees: @josteinaj I'm trying to rebase your patch onto the latest version of `px:html-outline` (I know, way overdue!) but I'm having trouble. The code now relies on the id attributes being present on section elements. I could try to solve it using temporary attributes, but this is a bit of hassle, so before getting in that I'd like to hear from you : why do you think it would be better to not modify the document?

Just FYI: the `px:html-outline` now a `heading-links-only` option. If I enable this option, in your test the first TOC item will point to the `h1` instead of the `body` (with generated ID). However the other items (the untitled ones) will then become `span` instead of `a` that link the `section`.

modules/common/html-utils/src/main/resources/xml/xslt/html5-outliner.xsl
bertfrees: This code has now moved to https://github.com/daisy/pipeline-modules/blob/3404b2828e70c2eab47c26d4e5e5b4de7248fdcb/common/html-utils/src/main/resources/xml/xslt/untitled-section-titles.xml (which also uses text nodes instead of strings).

modules/common/html-utils/src/main/resources/xml/xslt/html5-outliner.xsl
josteinaj: I don't quite remember what this was about, but I'll guess:

If you want to use the XSLT for building a navigation document or a ncc, you'll have to reference actual id attributes in the content documents. In a XProc script you could add id attributes beforehand and create a modified content document. But in the XSLT itself we can't really modify the content document, unless we do some `xsl:result-document` trickery to overwrite the content document (if that even works as expected when invoked from XProc).

bertfrees: > If you want to use the XSLT for building a navigation document or a ncc ...

OK but why would you want to use the XSLT directly, it's not a standalone XSLT? (Or, maybe it was, but it certainly isn't anymore.)

josteinaj: If this is only meant to be used through the XProc script, then I suppose modifying the original document is fine. Assuming that the XProc script that invokes `px:html-outline` uses the resulting modified document for further processing. Alternatively you could invoke a step that generates id attributes before invoking `px:html-outline`.

I don't depend on this XSLT or this XProc step anywhere currently so if you make it so that the original document is modified, that's fine by me.

bertfrees: OK.

josteinaj:
5 years since last activity. Closing.

PR #546, opened 2019-10-09
Some changes/improvements to the braille scripts

josteinaj:
- make px:html-to-pef.convert work the same way as px:dtbook-to-pef.convert
- add px:html-to-pef.store
- xml-to-pef: make it possible to preprocess before generating toc
- Support XSLT style sheets in stylesheet option of html-to-pef

### DIFF
--- a/html-to-pef/src/main/resources/xml/xproc/html-to-pef.store.xpl
+++ b/html-to-pef/src/main/resources/xml/xproc/html-to-pef.store.xpl
@@ -1,0 +1,45 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<p:declare-step type="px:html-to-pef.store" version="1.0"
+                xmlns:p="http://www.w3.org/ns/xproc"
+                xmlns:px="http://www.daisy.org/ns/pipeline/xproc"
+                exclude-inline-prefixes="#all"
+                name="main">
+    
+    <p:input port="html" primary="false"/>
+    <p:input port="obfl" primary="false">
+        <p:empty/>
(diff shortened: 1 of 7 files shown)


bertfrees:
- 9315de73211dda3e5a1b31795122bb7a530c7edf: make px:html-to-pef.convert work the same way as px:dtbook-to-pef.convert

   Needs [this fixup](https://github.com/daisy/pipeline-modules/commit/3a17685110a11e358c946f530b17540e08511aa4) but otherwise looks good.

- f07bb66290464511893a1a0a78504711a52779a0: add px:html-to-pef.store

  My intention is to go in the other direction, namely to remove as much as possible of the "load" and "store" steps in the script modules (as well as the `px:xml-to-pef.store` step). But I don't if this this refactoring is already for tomorrow, so I might accept the commit anyway because it increases the consistency between the different modules.

   By the way, something has gone wrong here because a file is in there twice. Must have been me during rebasing.

- 31dae9d61ec979b7fd9771382eb4ee45877e3568: Support XSLT style sheets in stylesheet option of html-to-pef

   Looks good

- c44f0bacba7fe4a9080cfe0069ab5536fb64d95d: xml-to-pef: make it possible to preprocess before generating toc

   Clever solution, but it is a tiny bit "magic" I think. I wonder if it would be more staightforward and easy to understand if there would simply be an option to disable the XSLT.

bertfrees:
Merged in c2fe0efd17d01f4ec82190e1ac60d87d7e37922c.

Issue #25, opened 2014-04-13
implement an EPUB3 OCF creator


ghost:
_From [rdeltour@gmail.com](https://code.google.com/u/rdeltour@gmail.com/) on April 27, 2011 17:44:46_
- Create the container manifest from file sets.
  - Add boilderplate files (container.xml, mimetype, etc)
  - ZIP the publication

_Original issue: http://code.google.com/p/daisy-pipeline/issues/detail?id=25_

PR #564, opened 2020-08-17
[New module] tts adapter for Google Cloud Text-To-Speech

NPavie:
# Google Text-To-Speech adapter module

We propose a new adapter to connect the DAISY pipeline 2 to Google Text-To-Speech using their REST api.

## Scheduler interface

A new "RequestSheduler" interface is proposed to delay requests execution. 
This interface is meant to be used to handle quotas and request retries.

A default implementation "ExponentialBackoffScheduler", following [Google recommandations](https://cloud.google.com/iot/docs/how-tos/exponential-backoff), is provided.

Those interface and implementation are added to the "tts-common" module.

## New properties

The following properties are used by the adapter : 
- org.daisy.pipeline.tts.google.apikey (mandatory) : API key to connect to Google Text-To-Speech services.
- org.daisy.pipeline.tts.google.samplerate (default to 22050) : sample rate of the audio output
- org.daisy.pipeline.tts.google.priority (default to 15) : priority of usage within the pipeline if other text engine are available

## Usage 

To use Google Cloud Text-to-Speech (https://cloud.google.com/text-to-speech/docs/quickstart-client-libraries) with this adapter in the pipeline, first go to the [Google cloud console](https://console.cloud.google.com)  :
1. In the project selection page, select or create a Cloud project.
2. Make sure billing is turned on for your Google Cloud project.
3. In the "API and Services" menu, open the "Library" panel, and search and activate the "Cloud Text-to-Speech API".
4. Still in In the "API and services" menu, open to the "Credentials" section and create an API key for this service.
5. Copy this API key in the pipeline configuration file, with the property "org.daisy.pipeline.tts.google.apikey".

### Notes 

This PR includes a merge commit that is the merge of our module private repo "merging" branch within the master of the pipeline repo.
(We are not planning to use this private repo for future work, but we would like to keep its history.)

### DIFF
--- a/assembly/NEWS.md
+++ b/assembly/NEWS.md
@@ -1,3 +1,10 @@
+Changes in release v1.13.6
+==========================
+
+### Modules
+
+- **NEW** Adapter for Google Cloud Text-To-Speech
+
 Changes in release v1.13.5
 ==========================
 
--- a/assembly/pom.xml
+++ b/assembly/pom.xml
@@ -1086,6 +1086,10 @@
                 </artifactItem>
                 <artifactItem>
                   <groupId>org.daisy.pipeline.modules</groupId>
+                  <artifactId>tts-adapter-google</artifactId>
+                </artifactItem>
+                <artifactItem>
+                  <groupId>org.daisy.pipeline.modules</groupId>
                   <artifactId>tts-adapter-osx</artifactId>
                 </artifactItem>
                 <artifactItem>
--- a/assembly/src/main/resources/etc/system.properties
+++ b/assembly/src/main/resources/etc/system.properties
@@ -116,6 +116,11 @@ org.daisy.pipeline.tts.config=${pipeline.config}/tts-default-config.xml
 #org.daisy.pipeline.tts.sapi.bytespersample=2
 #org.daisy.pipeline.tts.sapi.priority=7
 
+# Google Cloud TTS settings
+#org.daisy.pipeline.tts.google.apikey=
+#org.daisy.pipeline.tts.google.samplerate=22050
+#org.daisy.pipeline.tts.google.priority=15
+
 # Lame settings
 #org.daisy.pipeline.tts.lame.path=
 #org.daisy.pipeline.tts.lame.cli.options=
--- a/modules/.gitrepo
+++ b/modules/.gitrepo
@@ -7,5 +7,5 @@
 	remote = git@github.com:daisy/pipeline-modules.git
 	branch = master
 	commit = aec26b7fbcf3f72681e85e4807b98112c59cb505
-	parent = 2aac5a58056faa0d166fad11987f26ef43f4eb6e
+	parent = 9205a1931486e89b448126999174e612176bb3da
 	cmdver = 0.3.1
--- a/modules/bom/pom.xml
+++ b/modules/bom/pom.xml
@@ -434,6 +434,11 @@
       </dependency>
       <dependency>
         <groupId>org.daisy.pipeline.modules</groupId>
+        <artifactId>tts-adapter-google</artifactId>
+        <version>1.0.0-SNAPSHOT</version>
+      </dependency>
+      <dependency>
+        <groupId>org.daisy.pipeline.modules</groupId>
         <artifactId>tts-adapter-osx</artifactId>
         <version>3.0.4-SNAPSHOT</version>
       </dependency>
--- a/modules/tts/pom.xml
+++ b/modules/tts/pom.xml
@@ -34,6 +34,7 @@
     <module>tts-adapters/tts-adapter-osx</module>
     <module>tts-adapters/tts-adapter-sapinative</module>
     <module>tts-adapters/tts-adapter-qfrency</module>
+    <module>tts-adapters/tts-adapter-google</module>
     <module>zedai-to-ssml</module>
     <module>mathml-to-ssml</module>
   </modules>
--- a/modules/tts/tts-adapters/tts-adapter-google/README.md
+++ b/modules/tts/tts-adapters/tts-adapter-google/README.md
@@ -1,0 +1,21 @@
+# Google cloud Text-to-speech Adapter
+
+## New properties
+
+The following properties are used by the adapter : 
+- `org.daisy.pipeline.tts.google.apikey` (mandatory) : API key to connect to Google Text-To-Speech service.
+- `org.daisy.pipeline.tts.google.samplerate` (defaults to `22050`) : sample rate of the audio output
+- `org.daisy.pipeline.tts.google.priority` (defaults to `15`) : priority of usage within the pipeline if other text engine are available
+
+## Usage 
+
+To use Google Cloud Text-to-Speech (https://cloud.google.com/text-to-speech/docs/quickstart-client-libraries) with this adapter in the pipeline, first go to the [Google cloud console](https://console.cloud.google.com)  :
+1. In the project selection page, select or create a Cloud project.
+2. Make sure billing is turned on for your Google Cloud project.
+3. In the "API and Services" menu, open the "Library" panel, and search and activate the "Cloud Text-to-Speech API".
+4. Still in In the "API and services" menu, open to the "Credentials" section and create an API key for this service.
+5. Copy this API key in the pipeline configuration file, with the property "org.daisy.pipeline.tts.google.apikey".
+
+## Tests
+
+For unit test to work, please add the property "org.daisy.pipeline.tts.google.apikey" to either your configuration or pom, or add the option `-Dorg.daisy.pipeline.tts.google.apikey="your_key"` to your maven call. 
--- a/modules/tts/tts-adapters/tts-adapter-google/pom.xml
+++ b/modules/tts/tts-adapters/tts-adapter-google/pom.xml
@@ -1,0 +1,55 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+
+  <modelVersion>4.0.0</modelVersion>
+
+  <parent>
+    <groupId>org.daisy.pipeline.modules</groupId>
+    <artifactId>modules-parent</artifactId>
+    <version>1.13.7-SNAPSHOT</version>
+    <relativePath>../../../parent</relativePath>
+  </parent>
+
+  <version>1.0.0-SNAPSHOT</version>
+  <artifactId>tts-adapter-google</artifactId>
+  <packaging>bundle</packaging>
+
+  <name>DAISY Pipeline 2 module :: TTS Adapter for Google Cloud TTS</name>
+  <description>Implementation of the TTS API for the Google Cloud TTS</description>
+
+  <dependencies>
+    <dependency>
+      <groupId>org.daisy.pipeline.modules</groupId>
+      <artifactId>tts-common</artifactId>
+    </dependency>
+    <dependency>
+      <groupId>org.daisy.libs</groupId>
+      <artifactId>saxon-he</artifactId>
+    </dependency>
+  </dependencies>
+  
+  <properties>
+    <expose-services>
+      org.daisy.pipeline.tts.google.impl.GoogleTTSService
+    </expose-services>
+  </properties>
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.felix</groupId>
+        <artifactId>maven-bundle-plugin</artifactId>
+        <configuration>
+          <instructions>
+            <Import-Package>
+              net.sf.saxon.*;version="${saxon.versionRange}",
+              !org.daisy.common.spi,
+              *
+            </Import-Package>
+          </instructions>
+        </configuration>
+      </plugin>
+    </plugins>
+  </build>
+
+</project>
--- a/modules/tts/tts-adapters/tts-adapter-google/src/main/java/org/daisy/pipeline/tts/google/impl/GoogleRequestBuilder.java
+++ b/modules/tts/tts-adapters/tts-adapter-google/src/main/java/org/daisy/pipeline/tts/google/impl/GoogleRequestBuilder.java
@@ -1,0 +1,151 @@
+package org.daisy.pipeline.tts.google.impl;
+
+import java.util.HashMap;
+
+import org.daisy.pipeline.tts.rest.Request;
+import org.json.JSONObject;
+
+/**
+ * Google REST Request builder class.
+ *
+ * @author Louis Caille @ braillenet.org
+ */
+public class GoogleRequestBuilder {
+
+	/**
+	 * Google API key
+	 */
+	private String apiKey;
+
+	/**
+	 * Encoding sample rate (if null, uses the default encoding)
+	 */
+	private Integer sampleRate = null;
+
+	private GoogleRestAction action = GoogleRestAction.VOICES;
+	private String text = null;
+	private String languageCode = null;
+	private String voice = null;
+
+	/**
+	 * Create a new REST request builder for google services
+	 *
+	 * @param apiKey the API key to access google services (to create from your google cloud console API identifiers)
+	 */
+	public GoogleRequestBuilder(String apiKey) {
+		this.apiKey = apiKey;
+	}
+
+	/**
+	 * Set the action to execute
+	 *
+	 * @param action
+	 */
+	public GoogleRequestBuilder withAction(GoogleRestAction action) {
+		this.action = action;
+		return this;
+	}
+
+	/**
+	 * Mandatory - set the language code of the next requests
+	 *
+	 * @param languageCode language code of the voice
+	 */
+	public GoogleRequestBuilder withLanguageCode(String languageCode) {
+		this.languageCode = languageCode;
+		return this;
+	}
+
+	/**
+	 * Set the voice name to use for the next requests
+	 *
+	 * @param voice
+	 */
+	public GoogleRequestBuilder withVoice(String voice) {
+		this.voice = voice;
+		return this;
+	}
+
+	/**
+	 * Mandatory - Set the text to synthesise for the next requests
+	 *
+	 * @param text
+	 */
+	public GoogleRequestBuilder withText(String text) {
+		this.text = text;
+		return this;
+	}
+
+	/**
+	 * Set the sample rate of the next requests
+	 *
+	 * @param sampleRateHertz
+	 */
+	public GoogleRequestBuilder withSampleRate(int sampleRateHertz) {
+		this.sampleRate = Integer.valueOf(sampleRateHertz);
+		return this;
+	}
+
+	/**
+	 * Create a new builder instance with all value except the api key set to defaults.
+	 *
+	 * @return a new builder to use for building a request
+	 */
+	public GoogleRequestBuilder newRequest() {
+		return new GoogleRequestBuilder(apiKey);
+	}
+
+	/**
+	 * Build a rest request to send for google could text to speech
+	 *
+	 * @throws Exception
+	 */
+	public Request<JSONObject> build() throws Exception {
+
+		HashMap<String, String> headers = new HashMap<String, String>();
+		JSONObject parameters = null;
+
+		headers.put("Accept", "application/json");
+		headers.put("Content-Type", "application/json; utf-8");
+
+		switch(action) {
+		case VOICES:
+			// No specific parameters
+			break;
+		case SPEECH:
+			// speech synthesis errors handling
+			if (this.text == null || text.length() == 0) 
+				throw new Exception("Speech request without text.");
+			if (languageCode == null || languageCode.length() == 0)
+				throw new Exception("Language code definition is mandatory, please set one (speech request for " + text + ")");
+			
+			String jsonParameters =
+				"{"
+					+ "\"input\":{"
+						+ "\"ssml\":\"" + text + "\""
+					+ "},"
+					+ "\"voice\":{"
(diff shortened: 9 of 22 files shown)


bertfrees:
@louisCaille and @NPavie Great work!

Thanks for the nice summary. Is all of this documentation also present in the code?

- I see the README in tts-adapter-google, I'll update that with the version in your PR message because that seems a bit more detailed, OK?
- The things you wrote about the new ExponentialBackoffScheduler class, it would be nice if that is present in the Javadocs.
- It would also be great if the website is updated. The tts-common/doc/tts-config.md file is the source of the http://daisy.github.io/pipeline/modules/tts-common/doc/tts-config.html page, which explains how to configure TTS. Google should be mentioned here, and how to set it up.
- Javadocs for all new public methods: seems to be OK!

NPavie:
> @louisCaille and @NPavie Great work!
> 
> Thanks for the nice summary. Is all of this documentation also present in the code?
> 
> * I see the README in tts-adapter-google, I'll update that with the version in your PR message because that seems a bit more detailed, OK?

Yep it is ok for me

> * The things you wrote about the new ExponentialBackoffScheduler class, it would be nice if that is present in the Javadocs.

I'll update the javadoc.

> * It would also be great if the website is updated. The tts-common/doc/tts-config.md file is the source of the http://daisy.github.io/pipeline/modules/tts-common/doc/tts-config.html page, which explains how to configure TTS. Google should be mentioned here, and how to set it up.

Yep, forgot to update it, i'll do it right away

> * Javadocs for all new public methods: seems to be OK!

Cool :)

modules/tts/tts-utils/tts-common/src/main/java/org/daisy/pipeline/tts/ExponentialBackoffScheduler.java
```diff
@@ -0,0 +1,50 @@
+package org.daisy.pipeline.tts;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.UUID;
+
+public class ExponentialBackoffScheduler<RequestType> implements RequestScheduler<RequestType> {
+	
+	private int n = 0;
```
bertfrees, line 9: I find it not so intuitive that this number is never reset. It probably doesn't matter because for every invocation of `px:ssml-to-audio` a new `GoogleRestTTSEngine` is created with a new `ExponentialBackoffScheduler`, but it is not really clear that the class is supposed to be used like that. Maybe the documentation just needs some improvement.

In general, I think it is not super clear how the `RequestScheduler` interface is supposed to be used. More specifically, in what order are the methods supposed to be called. Would it be an idea to extract some of the logic from `GoogleRestTTSEngine` into the `org.daisy.pipeline.tts` package (tts-common) to make the interface more useful?

modules/tts/tts-utils/tts-adapter-google/src/main/java/org/daisy/pipeline/tts/google/impl/GoogleRestTTSEngine.java
```diff
@@ -0,0 +1,262 @@
...
+				// the answer is encoded in base 64, so it must be decoded
+				byte[] decodedBytes = Base64.getDecoder().decode(response.toString().substring(18, response.length()-2));
+
+				AudioBuffer b = bufferAllocator.allocateBuffer(decodedBytes.length);
+				b.data = decodedBytes;
+				result.add(b);
+				
+				isNotDone = false;
+
+			} catch (Throwable e1) {
```
bertfrees, line 139: I find this not so great. Isn't there a more specific exception we can catch when we get a 429 error?

modules/tts/tts-utils/tts-adapter-google/src/test/xprocspec/test_ssml-to-audio.xprocspec
```diff
@@ -0,0 +1,58 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<x:description xmlns:x="http://www.daisy.org/ns/xprocspec"
+               xmlns:px="http://www.daisy.org/ns/pipeline/xproc"
+               xmlns:ssml="http://www.w3.org/2001/10/synthesis"
+               xmlns:c="http://www.w3.org/ns/xproc-step"
+               xmlns:d="http://www.daisy.org/ns/pipeline/data"
+               script="http://www.daisy.org/pipeline/modules/ssml-to-audio/library.xpl">
+
+	<x:scenario label="synthesize with espeak">
```
bertfrees, line 9: This test is not run. It is a left-over from copying the espeak project. I'll fix it.

modules/tts/tts-utils/tts-adapter-google/src/test/java/org/daisy/pipeline/tts/google/impl/GoogleTTSSSMLTest.java
```diff
@@ -0,0 +1,137 @@
...
+import org.daisy.common.xslt.ThreadUnsafeXslTransformer;
+import org.daisy.common.xslt.XslTransformCompiler;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.Test;
+import org.xml.sax.SAXException;
+
+import com.xmlcalabash.util.TreeWriter;
+
+public class GoogleTTSSSMLTest {
```
bertfrees, line 26: Not sure what to do with this file. It hasn't been touched after it was copied from the espeak project.

modules/tts/tts-utils/tts-adapter-google/pom.xml
```diff
@@ -0,0 +1,81 @@
...
+    </dependency>
+    <dependency>
+      <groupId>org.daisy.pipeline.modules</groupId>
+      <artifactId>audio-encoder-lame</artifactId>
+      <scope>test</scope>
+    </dependency>
+    <dependency>
+      <groupId>org.daisy.libs</groupId>
+      <artifactId>saxon-he</artifactId>
+    </dependency>
```
bertfrees, line 41: Should be moved up.

modules/tts/tts-utils/tts-common/src/main/java/org/daisy/pipeline/tts/ExponentialBackoffScheduler.java
NPavie: You are right, i missed that >< 
Indeed the counter is meant to be reset if a request is successfull after the last delay.

I realise my scheduler interface needs some improvements, like some intel about retries and/or more functions like cancel and resume to respectively cancel a request or remove/reset the delay on a specified (or all) request(s)
I will update the interface and the javadoc to better reflect the idea of what the scheduler could do.

For the RequestScheduler, it is meant to delay the execution of requests based on constraints to be defined (like if an error is raised on a request or if quotas are met).

For the logic behind the TTS for rest API, i imagine some of the logic can be shared, as the main changes are the parameters and the headers of each request, but i'm not sure yet.
Maybe some kind of `RestRequest` interface could be defined and use in a `RestTTSEngine` with the `RequestScheduler` (if this is what you are thinking of)

modules/tts/tts-utils/tts-adapter-google/src/main/java/org/daisy/pipeline/tts/google/impl/GoogleRestTTSEngine.java
NPavie: it is the more common error code expected (quotas are exceeded).
i'll have to check the doc of google for the other possible code.

modules/tts/tts-utils/tts-adapter-google/src/test/java/org/daisy/pipeline/tts/google/impl/GoogleTTSSSMLTest.java
NPavie: I was thinking of keeping it in case of modifications of the transform to ensure nothing breaks (if modifications are required) when building the module independently.

If you think this part is not necessary, it can be deleted.

modules/tts/tts-utils/tts-adapter-google/pom.xml
NPavie: You mean in the parent pom ?

modules/tts/tts-utils/tts-adapter-google/src/main/java/org/daisy/pipeline/tts/google/impl/GoogleRestTTSEngine.java
bertfrees: I mean, this code expects that when you get a 429, an exception is thrown somewhere in the try block. So why not catch specifically that exception?

modules/tts/tts-utils/tts-adapter-google/src/test/java/org/daisy/pipeline/tts/google/impl/GoogleTTSSSMLTest.java
bertfrees: Well it could be useful of course: any test could possibly flag a regression. But I'd rather have tests that were added for a specific reason. This test really just seemed like a copy, and in general I don't like copying stuff because that way you potentially copy bugs and bad code. (This test in particular I don't like so much because it makes use of the Saxon and Calabash APIs, while I'd like to become independent of those APIs as much as possible.)

modules/tts/tts-utils/tts-adapter-google/src/main/java/org/daisy/pipeline/tts/google/impl/GoogleRestTTSEngine.java
NPavie: haaa sorry misunderstood.
I think it is due to the scheduler interface that can pause the thread (and so can throw an InterruptedException that needs to be catched).
I definitely need to rework the scheduler interface, i'm working on a proposal based on your previous comment to export more of the logic for the different rest engines to the tts-common.

modules/tts/tts-utils/tts-common/src/main/java/org/daisy/pipeline/tts/ExponentialBackoffScheduler.java
bertfrees: I was thinking of something more low level than that. Nothing to do with TTS or REST.

What I had in mind was just a way to let a scheduler execute certain actions. Basically you feed it "runnables" that may throw an exception when the action failed, in which case they are automatically rescheduled.

A very quick sketch to make the idea clear:

```java
interface Action {
    public void apply() throws RecoverableError, FatalError;
}

void process(RequestScheduler<Action>, Action) throws FatalError {
    ...
}

ExponentialBackoffScheduler<Action> scheduler = new ExponentialBackoffScheduler<>();

GoogleRestRequest speechRequest = ...;

process(
    scheduler,
    () -> {
        try {
            BufferedReader br = new BufferedReader(new InputStreamReader(speechRequest.send(), "utf-8"));
            StringBuilder response = new StringBuilder();
            String inputLine;
            while ((inputLine = br.readLine()) != null)
                response.append(inputLine.trim());
            br.close();
            byte[] decodedBytes = Base64.getDecoder().decode(response.toString().substring(18, response.length()-2));
            AudioBuffer b = bufferAllocator.allocateBuffer(decodedBytes.length);
            b.data = decodedBytes;
            result.add(b);
        } catch (Throwable e1) {
            try {
                if (request.getConnection().getResponseCode() == 429)
                    throw new RecoverableError();
                else {
                    SoundUtil.cancelFootPrint(result, bufferAllocator);
                    StringWriter sw = new StringWriter();
                    e1.printStackTrace(new PrintWriter(sw));
                    throw new FatalError(new SynthesisException(e1));
                }
            } catch (Exception e2) {
                SoundUtil.cancelFootPrint(result, bufferAllocator);
                StringWriter sw = new StringWriter();
                e2.printStackTrace(new PrintWriter(sw));
                throw new FatalError(new SynthesisException(e2));
            }
        }
    }
);
```

modules/tts/tts-utils/tts-adapter-google/pom.xml
bertfrees: I mean it's not a test dependency :-). Don't worry I can fix this.

modules/tts/tts-utils/tts-common/src/main/java/org/daisy/pipeline/tts/ExponentialBackoffScheduler.java
NPavie: Ok i'll work on that

modules/tts/tts-utils/tts-adapter-google/src/main/java/org/daisy/pipeline/tts/google/impl/GoogleRestTTSEngine.java
```diff
@@ -0,0 +1,226 @@
...
+						} else {
+							throw new FatalError(e);
+						}
+					} catch (IOException responseCodeError) {
+						throw new FatalError("could not retrieve response code for request", responseCodeError);
+					}
+				}
+			});
+			
+			mRequestScheduler.launch(requestUuid);
```
bertfrees, line 138: Is there any reason why you would not "launch" the request immediately after adding it? If "add" and "launch" are combined there is no need for the UUID.

NPavie: I keep a fix storage process for all scheduler (with the add function implemented in the abstract scheduler class), and let futur implementations of the scheduler class decide what to do when launching a specific request with that approach (or use existing scheduler.
if you think its a problem i can change that

bertfrees: Not sure I get it.

NPavie:
I think i'm good with the modification requested : 
- Scheduler interface has been simplified following recommendations (and implementations using it have been updated accordingly)
- The unused xprocspec test (and xproc file) has been removed along dependencies
- The javadoc and readme has been updated
- the tts-config documentation has been updated

I also added a small generalisation on the Request class to make it a template; with the request content type now being a class parameter (in case someone would like to use something else than a JSONObject as content structure to be sent with a request).

I had to do an edited merge due to the changes in the modules repo structure and to resync the branch with master's latest updates.

Is there anything else preventing the merge in master ?
(Should i moved the pull request to the modules repo instead ?)

bertfrees:
Don't worry, I'm moving it to the modules repo. I combined/split some commits, cleaned some commit messages and added some of my own commits.

Check it out at https://github.com/daisy/pipeline-modules/pull/25.

Just one remark: next time try to avoid merge commits on the super repo level, it makes things a bit more complicated.

There is at least one thing left to do before I merge: it's the capturing of the 429 error that I had a comment about earlier. The code is better than before, but there is still room for improvement. If the response code is 429, is the exception a `IOException` or a `MemoryException`, or can it be both? Also, if you know the exception will be thrown in the `speechRequest.send()` call, it is better to catch it on that specific line.

NPavie:
> Just one remark: next time try to avoid merge commits on the super repo level, it makes things a bit more complicated.

I was forced to do a complicated merge due to a change in the modules repo structure that prevented the branch rebase/fast-forwarding and was creating many conflicts when cherry picking of commits.
(the changes would have also made the PR merge difficult as i understand, or at least git and github was telling so on my repo)
Next time i'll work directly on lower repos level, it seems easier that way, and less prone to other repos changes creating divergences between branches being reviewed for PR on the pipeline repo.
(I'm thinking git subrepo makes it not ideal to do PR on the pipeline repo so i may stop doing PR on it for next modules)
 
> There is at least one thing left to do before I merge: it's the capturing of the 429 error that I had a comment about earlier. The code is better than before, but there is still room for improvement. If the response code is 429, is the exception a `IOException` or a `MemoryException`, or can it be both? 

There is always room for improvement in every code, so i may need more specifics for that case.

I don't know why but my Eclispe recommended to merge catch blocks, i will split them again. 
(i do prefer splitting them too for readability, i though it was something like a new recommendation for java that was integrated in eclipse, maybe some kind of performance improvement recommendation by doing less catch ?).

On the other hand, I don't think it is a big deal in that case to specifically know which type of exception (IO or memory) it is : 
in any case, if there is a different response code, a different IO problem or a Memory error while allocating a buffer, it means a fatal error should be raised with the exception sent back along the FatalError. 
It becomes relevant if you want to add a custom messages with the exceptions that are not related to the request error code.
Specializing the FatalError exceptions based on error codes messages could be more of use for error reports with remote services, but the 429 code is the only one i find that can be recoverable and is mandatory to process for relaunch. 
The error messages specialization (localization ?) for other error codes can always be added on the way as an improvement though.

I will do as you request, but i' have to say i'm starting to think we are focusing on reworking the code style, more than improving the actual logic and/or performance of the code.
This could delay the PR merge indefinetly.

> Also, if you know the exception will be thrown in the `speechRequest.send()` call, it is better to catch it on that specific line.

Why is that ? 
if you are talking readability, its more of a code style point of view (i.e. i prefer to centralize error recovery at the end of a block, so the expected execution logic of the block is complete before the "exceptional/on errors" logic is displayed).
If you are talking performance, i'm not sure it is of great change to do smaller try/catch.
It may even be the opposite : i always hear from senior devs and for the majority of languages, including java, that less try catch is better for performance, as switching to exception handling context always comes with a significant cost on memory and execution heap for processors. 
If you have a performance impact analysis report on that code pattern for java that says it is better that way, i'd be glad to read it.

bertfrees:
> I was forced to do a complicated merge due to a change in the modules repo structure that prevented the branch rebase/fast-forwarding and was creating many conflicts when cherry picking of commits.

You don't always have to keep up with the latest master, unless I specifically ask for it. In that case, because I indeed made a change to the folder structure, I would definitely not have asked you to rebase.

But the problem already started with the first merge commit, where you added Louis' branch. This should either has been a rebase (on the super repo level), or it should have been done on the level of the pipeline-modules repo.

> I'm thinking git subrepo makes it not ideal to do PR on the pipeline repo so i may stop doing PR on it for next modules.

It is also really useful when you make changes to multiple repos, and as long as you stick to the no-merging rule, there is really no issue with it.

bertfrees:
That's right, it's a code style - or rather code hygiene - issue. But it's important! People can have different preferences regarding style, but one should not compromise on things like readability.

> I don't know why but my Eclispe recommended to merge catch blocks. [...]

I'm not talking about combining multiple types of exception in a single catch, like the `catch (IOException | MemoryException e)`, per se. It's perfectly fine to treat multiple exceptions the same way if that's appropriate for the situation.

I'm talking specifically about the 429 exception here. This is indeed about readability. It is hard to follow the program flow. What has exactly happened before you end up inside the `if (speechRequest.getConnection().getResponseCode() == 429)` condition? How did you get there? This is generally a problem with big try blocks, but even more so in this specific case because the error information comes from the `speechRequest.getConnection()`, not from an exception object.

Regarding performance: yes you could be right, I'm not really aware of the performance impact of these things. But smaller try blocks don't *necessarily* result in more switching to exception handling context, do they? That depends on the nesting, right?

NPavie:
For the merge aspect, i may have forgot to say that at first i tried to do a rebase but as Louis started from an empty simple module repo, the rebase was asking me lots and lots of validations during the rebase process (and it was taking a very long time).
I found on other forums and tutorials that for that particular case (merge of a source repo branch inside a different destination repo branch while keeping the source repo history) it was recommended to update the source repo directories structure to match the destination repo, and then to do a merge in the destination repo (thus keeping the commit history accurate and getting the correct result).
I found other examples that used rebase but were completely rewriting the tree history, and i was not confident about that.
(i may not master git as much as you though, so maybe i missed something while trying to merge the repos with the rebase command, i may retry for other adapters).

On the code hygiene side, you are the lead on the pipeline side so i'll do what you ask, but i have to warn you : 
You are making this PR very long to deploy based on (in my opinion) over-estimated readability difficulties, especially compared to other critical parts of the pipeline code.
Sure the hygiene side is of importance, but that seems exaggerating in that case :
we are talking about a less than 15 lines try block, with very few throwing cause. 
For the catch block, the `speechRequest.getConnection().getResponseCode() == 429` seems quite self explanatory, as to tell we are checking what the server response code was with the speech request connexion when an exception is raised.

bertfrees:
> it was recommended to update the source repo directories structure to match the destination repo, and then to do a merge in the destination repo (thus keeping the commit history accurate and getting the correct result).

Sure, but not on the super repo level, that's a rule. But really, don't worry about it, it was not a big issue. I just mentioned it for next time.

> you are the lead on the pipeline side so i'll do what you ask

Haha, no, I'd rather have that you agree with me. I don't want to be a dictator!

> You are making this PR very long to deploy based on (in my opinion) over-estimated readability difficulties, especially compared to other critical parts of the pipeline code.

I'm sorry, the argument about other parts of the Pipeline code doesn't hold. The existing code is the way it is, but for new code I think we can have more stringent requirements.

If this PR is taking very long it can't be just because of this issue, because this is very quick to fix. The other changes that I requested before were more involved, but also worth it (I hope you agree).

Anyway, this is obviously bugging me much more than you, so maybe it's just me. So I suggest I'll tackle it myself. I don't want to ask you things you're not convinced of.

bertfrees:
This will also give me the opportunity to try to run the tests with an API key.

bertfrees:
> This will also give me the opportunity to try to run the tests with an API key.

Hmm, so I can't try it out without linking a billing account? Too bad.

NPavie:
> > This will also give me the opportunity to try to run the tests with an API key.
> 
> Hmm, so I can't try it out without linking a billing account? Too bad.

Yeah that's one of the problem with Google and Amazon services : you need to fill a payment method to use their API.

(IBM has a free "Lite" plan that you can use without adding billing information.)

bertfrees:
Merged in https://github.com/daisy/pipeline-modules/commit/0d51e487ce286b3ed59f5794e482ab42c27ba627

PR #601, opened 2021-06-02
Change default for braille in epub enhancement to false.

kalaspuffar:
Hi @bertfrees 

From the discussion, we had yesterday the braille conversion should not be default for the epub3 enhancement. So I was told to make a PR to both repositories so this change doesn't make a divergence for the MTM branch.

Best regards
Daniel


bertfrees:
This was fixed in https://github.com/daisy/pipeline-modules/commit/a733fec45138a4132019c02e38a2de8c43d83814

PR #620, opened 2022-01-19
Webui login

kalaspuffar:
Hi @bertfrees 

I've been working on helping @josteinaj  to set up login flow for their pipeline installation. We ran into a wall when you could not set up an API connection with authid and secret key and I guess it is because it's not using an URL encoded Base64 string for signing. 

This PR is trying to resolve this issue. I've separated it into two different commits. The first one was all changes that were required for me in order to build the Pipeline webservice and webui. Some SNAPSHOT version changes and also I changed to the new URI for restlet required of some packages.

More over I added the feature to create a zip file of the webui package. Not required so I can remove that change if you like.

The next commit is all code changes. Mainly I've changed to use the Base64.getUrlEncoder().withoutPadding().encode() for the hash that should be compared, this makes the process more reliable. I also had to change the UI component verifying the login to use a known endpoint like `/jobs` using the default root it never ran any login flow and always returned 404 so the webui could never have worked with authid and secret key.

I hope that these changes are OK, I understand that changing the API could be bad if a lot of people have implemented it against the old one and if so we could gradually change or just implement the change with the endpoint to ensure that the webui works. Then again @josteinaj could not reliably get the API to work.

Best regards
Daniel

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -20,3 +20,5 @@ bin/
 target/
 /build/
 /node-modules/
+*.iml
+.idea
--- a/Makefile
+++ b/Makefile
@@ -75,6 +75,14 @@ dist-docker-image : assembly/.compile-dependencies | .maven-init .group-eval
 .PHONY : dist-cli-deb
 dist-cli-deb : cli-$(cli/VERSION)-linux_386.deb
 
+.PHONY : dist-webui-zip
+dist-webui-zip : assembly/.compile-dependencies
+	# see webui README for instructions on how to make a signed package for distribution
+	cd webui && \
+	./activator -Dmvn.settings.localRepository="file:$(CURDIR)/$(MVN_WORKSPACE)" clean universal:packageBin | $(MVN_LOG)
+	mv webui/target/universal/*zip .
+
+
 .PHONY : dist-webui-deb
 dist-webui-deb : assembly/.compile-dependencies
 	# see webui README for instructions on how to make a signed package for distribution
--- a/assembly/pom.xml
+++ b/assembly/pom.xml
@@ -83,7 +83,7 @@
       <dependency>
         <groupId>org.daisy.pipeline</groupId>
         <artifactId>framework-bom</artifactId>
-        <version>1.14.4</version>
+        <version>1.14.5-SNAPSHOT</version>
         <type>pom</type>
         <scope>import</scope>
       </dependency>
@@ -2765,7 +2765,7 @@
     <repository>
       <id>restlet-repo</id>
       <name>Public online Restlet repository</name>
-      <url>http://maven.restlet.org</url>
+      <url>https://maven.restlet.talend.com</url>
       <releases>
         <enabled>true</enabled>
       </releases>
--- a/clientlib/java/clientlib-java-httpclient/src/main/java/org/daisy/pipeline/client/http/Pipeline2HttpClient.java
+++ b/clientlib/java/clientlib-java-httpclient/src/main/java/org/daisy/pipeline/client/http/Pipeline2HttpClient.java
@@ -252,20 +252,8 @@ public class Pipeline2HttpClient {
 
 			url += "authid="+username + "&time="+time + "&nonce="+nonce;
 
-			String hash = "";
 			try {
-				hash = calculateRFC2104HMAC(url, secret);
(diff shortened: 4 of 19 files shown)


kalaspuffar:
Hi @bertfrees 

I want to learn how to bump the version in the future, so starting the work on any components would be easier. 

I've now implemented an extra URL parameter called `urlsign`, which is the URL encoded version of the hash.

So now we could handle both versions, and if you have a better suggestion of the name, I could change it. I've tested this with both the WebUI and the command-line tool, and it works with both. The CLI version I tested with had not been updated, so it should have been running with the old implementation.

Best regards
Daniel

kalaspuffar:
Hi @bertfrees 

I added some documentation also so the users could choose this alternative implementation option. But, of course, if we want to change going forward, we could replace the original documentation and gradually move over and deprecate the old implementation. But that is up to the project to decide.

Best regards
Daniel

bertfrees:
> I want to learn how to bump the version in the future, so starting the work on any components would be easier.

The version of pipeline-framework that was committed to the super project was https://github.com/daisy/pipeline-framework/commit/25e35ee55cfb4a83351c3db97245a175daa30834. To bump the version I would commit https://github.com/daisy/pipeline-framework/commit/0176e0532e2613c4db9b35dbb5eef14ab57baaea. To do this I would run the command:

```sh
source .git-utils/git-subrepo/.rc
git subrepo commit framework 0176e05
```

The solution with the two parameters is definitely better because we don't change the API. Also thanks for the documentation. That clarified some things for me.

But I'm still not sure why all this is needed. Why is the web UI unable to pass the hash string (with the padding etc.)?

@rdeltour What do you think of this?

kalaspuffar:
Hi @bertfrees 

Well, there are multiple reasons why I think this is an improvement. First, if you tell people to do encoding that then could automatically be decoded in some systems then they need to double encode and so on which never is a good solution. Changing to a defined standard it's easier to implement and hopefully will introduce fewer issues. 

https://datatracker.ietf.org/doc/html/rfc4648#section-5

I built a client a couple of years ago and this standard was chosen to solve a lot of issues that we ran into with different browsers and operating systems.

And when it comes to the client we need to change it anyway because the encoding was not complete.

```
				for (int i = 0; i < hash.length(); i++) {
					// Base64 encoding uses + which we have to encode in URL parameters.
					// Hoping this for loop is more efficient than the equivalent replace("\\+","%2B") regex.
					c = hash.charAt(i);
					if (c == '+') hashEscaped += "%2B";
					else hashEscaped += c;
				} 
```

This function only handles one of the issues. Slash and equal signs are interpreted differently by different parsers or query builders. So if we want to use the old format then all of them should be encoded.

Best regards
Daniel

bertfrees:
> https://datatracker.ietf.org/doc/html/rfc4648#section-5

Thanks for the pointer. Yes, I agree one encoding step less to do is less things that can go wrong and it would've been better if we had gone for that solution from the beginning.

> And when it comes to the client we need to change it anyway because the encoding was not complete.

> This function only handles one of the issues. Slash and equal signs are interpreted differently by different parsers or query builders. So if we want to use the old format then all of them should be encoded.

So there's the bug! Thanks, this is what I wanted to know. I needed to know whether there is a problem in the client that can be fixed, or whether there is a fundamental problem with the web API. To be completely reassured, I'd like to first check if fixing this bug fixes the problem and is not too hard. In a second phase we can consider improving the API by introducing a new "urlsign" parameter. At first sight fixing the bug looks pretty straightforward. A few lines higher up in the code, `URLEncoder.encode()` is used to encode all the other query parameters. I wonder why the same technique is not used below to encode the "sign" parameter.

```java

kalaspuffar:
Hi @bertfrees 

Not sure if the solution is complete, but the changes seem reasonable.

Best regards
Daniel

josteinaj:
It seems that orgaisy on docker hub does not build the webui automatically. Not a problem, I just assumed it were. In any case, I built it manually, tagged it and pushed it, so that it's now available as the docker image `daisyorg/pipeline-webui:v2.7.1`. I tested locally that it works.

bertfrees:
For the "urlsign" related changes see this new PR: https://github.com/daisy/pipeline/pull/637

PR #625, opened 2022-02-21
Proposal : adding windows drives test for ConfigReader URI resolver

NPavie:
On windows system, the tts-common project is not passing the URITest.absolutePath test due to unix system root used for the test.

I propose the following changes to make the test path and extend absolute URI resolution for windows system :
- in URIInsideConfig method of ConfigReader  :
  - Checking for windows drives in the path
  - Adding a default mapping between unix root and windows system drive

URITest class is also changed to consider windows subsystem for the absolutPath test


bertfrees:
I have solved this a different way. The `URIinsideConfig` and `readFromURIinsideConfig` methods must be called with a non-null base URI argument. Will reference this PR from the commit that I will push later.

bertfrees:
I've replaced this by commit https://github.com/daisy/pipeline-modules/commit/fb253a4cb3ffd5f527c31de868095b2d84aea95e.

PR #626, opened 2022-02-23
tts-adapter-sapinative fix

NPavie:
Fixing the Windows SAPI adapter for the pipeline.

The dlls build process is being migrated from the sapi-nar package to the adapter, also including 
- new release DLLs distributed as resources in the repository, 
- a new VS2019 build solution and project to be used for DLL rebuild if needed
- A README to indicate requirements to install and how to build the dlls and the adapter

The following changes are done in the code:
- Improved voice data recovery in the library to also get locale, gender and age when available
- Replace unsupported ssml mark tag by sapi bookmark tag in transform-ssml xslt
- Commented ending-mark definition as SAPI does not detect bookmarks after the last viseme
- Used the voices locale, gender and age retrieved for the voices list created by the SAPIEngine
- Added matching enums between the java and the native side for SAPI errors identification

Unit tests have been updated accordingly.

Some investigation and additionnal changes might be required to fix a remaining issue:
On some occasions, the SAPINative tests are raising an exception when the DLL is being disposed after all unit tests have finished, but the build remains valid with the exception being displayed after tests have passed in success, and an error log is created when it occurs.
This should not have impact in production as long as the dll remains loaded in the JVM (that is, until the pipeline is exiting)


bertfrees:
Thanks very much!

> On some occasions, the SAPINative tests are raising an exception when the DLL is being disposed after all unit tests have finished, but the build remains valid with the exception being displayed after tests have passed in success, and an error log is created when it occurs.

I believe we agreed that there should be a warning about this in the test log?

NPavie:
> I believe we agreed that there should be a warning about this in the test log?

You are right i forgot ^^'
I'll add the notice in the dispose method of the SAPINativeTest class.

PR #632, opened 2022-03-28
Don't echo echo commands in the Makefile

egli:
Normally make always echos commands. If the command itself is an echo command it does not need to be echoed as the sole purpose of it is to print something.

The special built-in target `.SILENT` makes sure that for those rules that are just meant for printing the recipe is not printed also.

This makes sure that only the text from the echo statements is actually printed, instead of both the text and the echo commnd itself.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -120,6 +120,7 @@ run-gui : $(dev_launcher)
 	$< gui
 
 .PHONY : run-cli
+.SILENT : run-cli
 run-cli :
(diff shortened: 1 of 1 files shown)


bertfrees:
.make/main.mk already contains a `.SILENT` target without prerequisites, which silences everything:

``` Makefile
ifndef VERBOSE
.SILENT:
endif
```

Adding `.SILENT : run-cli` etc. ondoes this.

I pushed a fix: 84a1759348

PR #655, opened 2022-10-11
SAPI and Onecore Text-to-speech unified adapter

NPavie:
# SAPI and Onecore TTS support

Changes required for enabling Microsoft SAPI and OneCore text to speech synthesis within the DAISY pipeline 2.

Given the similarity between the 2 engines (OneCore and SAPI Voices uses the same structure but stored under different registry keys), connection to those are enabled through the tts-adapter-sapinative adapter, that manages the 2 native libraries using the different APIs needed to interact with them : 
- ~~the sapinative library has been fixed to correctly handle ssml, including marks, with the SAPI TTS runtime~~
  - ~~it has been moved from the sapi-nar project to the tts-adapter-sapinative one, as only this adapter uses it~~
- A new onecorenative library is hereby provided to synthesize speech from ssml markups with the OneCore TTS runtime
  - As for SAPI, the engine also handles marks
  - This library uses C++/WinRT bindings to interact with the OneCore runtime

Both libraries are provided pre-built as dlls within the adapter for windows x86 and x64 systems, and the correct dlls are automatically loaded by the service based on the JVM architecture.

If needed, they can be rebuild with visual studio 2022 community, using the sapi-onecore-native.sln solution.

Note for users of the adapters that have third party SAPI voices : 
voices that are installed under the 32bit (x86) registry cannot be used through this adapter if the jvm running is not also 32 bit.
This is due to x86 and x64 voices registry of SAPI being not unified, meaning that an 64bit application that uses SAPI will have only access to the voices that are installed in the 64bit registry (and similarily, a 32bit application will only have access to the 32 bits voices registry).
Some hacks can be found on the web to "copy" a voice from x86 to x64 registry, but this hack has been tested only for Microsoft Voices and might not work for a third party voice.
To use 32bit voices, you will need to run the pipeline with a 32bit java runtime.


bertfrees:
@NPavie Thanks!

> - the sapinative library has been fixed to correctly handle ssml, including marks, with the SAPI TTS runtime
>
>   - it has been moved from the sapi-nar project to the tts-adapter-sapinative one, as only this adapter uses it

If I'm not mistaken this was already done in commits https://github.com/daisy/pipeline-modules/commit/0eb76a386cf5e69df8cb20d5d7d6648ce15d6fe2 and https://github.com/daisy/pipeline-modules/commit/d2cad09325d21131805775113a196dfb297f1c2b. So it looks like at least the first three commits in this PR are not needed. Could you please check and update the PR accordingly?

bertfrees:
The "fixup" commits f7c50407fe04fbc65b3fce17e956c9c1458e959e, 26b708aaa578b3336af16c77b801d35ff881177c and b5c49e2d6b39b61ec827adc7ded2a0a372633eaf should be combined with the right commits that they fix.

modules/tts/tts-adapters/tts-adapter-sapinative/src/main/resources/transform-ssml-onecore.xsl
```diff
@@ -0,0 +1,38 @@
+<xsl:stylesheet xmlns:xsl="http://www.w3.org/1999/XSL/Transform"
```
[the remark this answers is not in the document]
bertfrees: Actually, it seems that the original XSLT is not used anymore because you found a better fix for the SSML marks issue.

modules/tts/tts-adapters/tts-adapter-sapinative/src/main/java/org/daisy/pipeline/tts/onecore/impl/OnecoreEngine.java
```diff
@@ -259,8 +263,7 @@ public Collection<Voice> getAvailableVoices() {
 					original.getGender().get()
 				)
 			);
-
-			Logger.info("Voice available : " + original.toString());
+			//Logger.info("Voice available : " + original.toString());
```
bertfrees, line 266: Either remove, leave as is, or change log level. I don't like if it's commented out.

modules/tts/tts-adapters/tts-adapter-sapinative/.gitignore
```diff
@@ -1,2 +1,10 @@
...
\ No newline at end of file
+*.log
+*.vcxproj.*
+src/main/resources/x64/*
+!src/main/resources/x64/sapinative.dll
+!src/main/resources/x64/onecorenative.dll
+src/main/resources/x86/*
+!src/main/resources/x86/sapinative.dll
+!src/main/resources/x86/onecorenative.dll
+build/
```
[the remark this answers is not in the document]
NPavie:
```
*.vcxproj.* -> exclude files generated by visual studio for user preferences
src/main/resources/x64/* -> exclude all generation files that are not the library (can be debug symbols and static libs that could be used for linking in other C++ project)
!src/main/resources/x64/sapinative.dll -> only keep the dll 
!src/main/resources/x64/onecorenative.dll -> only keep the dll
src/main/resources/x86/* -> same as x64 folder
!src/main/resources/x86/sapinative.dll -> same as for x64 folder
!src/main/resources/x86/onecorenative.dll -> same as for x64 folder
build/ -> exclude intermediate build folder of the VS solution
```

bertfrees:
@NPavie I'm doing some cleanup right now. I have reduced the PR to three commits. I still have some work to make each commit compile and to clean up the source code a bit. I also plan to delete SAPIService and rename OnecoreService to SAPIService.

modules/tts/tts-adapters/tts-adapter-sapinative/src/main/java/org/daisy/pipeline/tts/onecore/impl/OnecoreEngine.java
bertfrees: I've removed it.

modules/tts/tts-adapters/tts-adapter-sapinative/src/main/resources/transform-ssml-onecore.xsl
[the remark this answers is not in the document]
bertfrees: I have dropped the legacy SAPIservice.

modules/tts/tts-adapters/tts-adapter-sapinative/pom.xml
```diff
@@ -44,26 +44,26 @@
   </dependencies>
 
   <properties>
+    <!--org.daisy.pipeline.tts.sapi.impl.SAPIservice,-->
```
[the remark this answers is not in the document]
bertfrees: I have dropped the legacy SAPIservice.

bertfrees:
@NPavie Everything is in a single commit now. There is now a single TTSService called `SAPIService` (in the `org.daisy.pipeline.tts.sapi.impl` package) for both legacy SAPI and OneCore voices.

Ideally I'd like to move all Java classes to the `org.daisy.pipeline.tts.sapi.impl` package because we don't need public packages. How much work is that? If it is too cumbersome we can leave it.

I have some more questions, I will add comments in the code.

modules/tts/tts-adapters/tts-adapter-sapinative/.gitignore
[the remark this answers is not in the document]
bertfrees: Thanks. So it's all from Visual Studio. `/packages/` too. What about `*.log` and `./.apt*`?

modules/tts/tts-adapters/tts-adapter-sapinative/sapinative.sln
```diff
@@ -1,9 +1,16 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.32106.194
+# Visual Studio Version 17
```
bertfrees, line 3: You updated to Visual Studio 17 and in the commit message you say:

> updated the projects to use Visual Studio 2022 and updated tool chain

but the README says "Visual studio 2019 or newer". Is that still correct or does it need updating?

modules/tts/tts-adapters/tts-adapter-sapinative/src/main/java/org/daisy/pipeline/tts/sapi/impl/SAPIService.java
```diff
@@ -0,0 +1,173 @@
...
+					Onecore.dispose();
+					throw new SynthesisException(
+							"Onecore initialization failed with error code '" + res + "': " + OnecoreResult.valueOfCode(res));
+				}
+			} catch (Exception e) {
+				Logger.warn(e.getMessage());
+			}
+			try {
+				int sapiSampleRate = convertToInt(params, "org.daisy.pipeline.tts.sapi.samplerate", 22050);
+				int sapiBytesPerSample = convertToInt(params, "org.daisy.pipeline.tts.sapi.bytespersample", 2);
```
bertfrees, line 64: These settings are used for legacy SAPI voices, but not for OneCore voices. Why was this again? (You possibly already explained but I forget.) Should this be mentioned in the [documentation](https://github.com/daisy/pipeline-modules/blob/master/tts/tts-common/doc/index.md#sapi)?

modules/tts/tts-adapters/tts-adapter-sapinative/src/test/java/ignore
```diff
@@ -1 +1,7 @@
-org/daisy/pipeline/tts/sapinative/impl/SAPINativeTest.java
+# Tests are not run by default. It only makes sense to run them on Windows.
+# Don't enable OnecoreTest and SAPITest when SAPIServiceTest is
+# also enabled. Due to concurrent attempt to load or dispose of
+# library, we cannot run them concurrently.
```
bertfrees, line 4: Can you elaborate a bit on this? What exactly is the problem? Is there are way to overcome it?

modules/tts/tts-adapters/tts-adapter-sapinative/src/test/java/org/daisy/pipeline/tts/sapi/impl/OnecoreTest.java
```diff
@@ -0,0 +1,240 @@
...
+		Collection<Voice> voices = allocateEngine().getAvailableVoices();
+		Assert.assertTrue(voices.size() > 1);
+	}
+
+	// https://github.com/microsoft/Windows-universal-samples/tree/main/Samples/SpeechRecognitionAndSynthesis
+	// <speak version='1.0' xmlns='http://www.w3.org/2001/10/synthesis'
+	// xmlns:xsi='http://www.w3.org/2001/XMLSchema-instance'
+	// xsi:schemaLocation='http://www.w3.org/2001/10/synthesis  http://www.w3.org/TR/speech-synthesis/synthesis.xsd'
+	//"<speak version='1.0' xmlns='http://www.w3.org/2001/10/synthesis'><s>" + x + "</s></speak>";
+	// <ssml:speak xmlns:ssml="http://www.w3.org/2001/10/synthesis" version="1.0"><s:s xmlns:tmp="http://" xmlns:s="http://www.w3.org/2001/10/synthesis" id="s1">small sentence</s:s><ssml:break time="250ms"/></ssml:speak>
```
bertfrees, line 79: What is this?

modules/tts/tts-adapters/tts-adapter-sapinative/sapinative.sln
NPavie: (sorry i was on the move for AVH yesterday)
You are right i need to change this to Visual studio 2022
For the version in the sln file, i think 2022 is the version 17 of visual studio.

i'll do a pass on the code this afternoon.

modules/tts/tts-adapters/tts-adapter-sapinative/src/main/java/org/daisy/pipeline/tts/sapi/impl/SAPIService.java
NPavie: For OneCore voice, you cannot change the sample rate and bytes per sample settings of the synthesizer anymore.
(Or rather I have not found any settings that would allow to change them on the WinRT API pages or from the documentation exposed when working in visual studio with c++/winrt)

It might be indeed useful to mention this in the documentation that list the options for TTS engines.

modules/tts/tts-adapters/tts-adapter-sapinative/src/test/java/ignore
NPavie: I don't remember exactly but if i recall, when launching the test simultaneously, it would try to load the dll multiple times and it would make the tests failed.

I'll need to retest that to exactly know what was going on on the jvm side, if it was a runtime exeception due to dll reloading or if it was something linked to the libs initialization process that would crash the rest of the code (possibly due to reallocation of global variables requested in the tests initialization).

modules/tts/tts-adapters/tts-adapter-sapinative/src/test/java/org/daisy/pipeline/tts/sapi/impl/OnecoreTest.java
NPavie: Oups forgot to clean that up.
I was keeping traces of the ssml used in MS documentation to check what was wrong with our ssml.
I'll remove that.

bertfrees: OK cool.

modules/tts/tts-adapters/tts-adapter-sapinative/src/test/java/ignore
bertfrees: Or was it an issue due to parallel execution of tests? (because you mention "concurrent")

NPavie:
A quick update with the latest commit update : 
- I change the way the libraries initialization and disposable are handle by the service, to allow for multiple class to interact with it, so that tests class can now run concurrently : 
  - 2 set of "load" and "release" methods are now provided by the service, with a "request counter" for each library to only dispose the library when every objects requesting it (with the load method) have called the release method at the end
  - Tests have been changed to use those load and release methods instead of the previous "loadDLL", "initialize" and "dispose" calls. Tests run concurrently and report no failures on my windows machine. 
- I re enable the ending mark test and with the ssml fix, the ending marks are correctly reported.

modules/tts/tts-adapters/tts-adapter-sapinative/src/test/java/org/daisy/pipeline/tts/onecore/impl/OnecoreTest.java
```diff
@@ -0,0 +1,274 @@
...
+	// 	String bookmark = "endingmark";
+	// 	long connection = speakCycle(SSML("this is an ending mark <bookmark mark=\"" + bookmark
+	// 	        + "\"/> "));
+	// 	String[] names = OnecoreLib.getBookmarkNames(connection);
+	// 	long[] pos = OnecoreLib.getBookmarkPositions(connection);
+	// 	OnecoreLib.closeConnection(connection);
+	// 	Assert.assertSame(1, names.length);
+	// 	Assert.assertSame(1, pos.length);
+	// 	Assert.assertEquals(bookmark, names[0]);
+	// }
```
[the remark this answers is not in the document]
bertfrees: Fixed.

bertfrees:
Perfect, thanks! You can just add fixup commits to the branch, that makes it easier for me what changed. I can then squash afterwards.

modules/tts/tts-adapters/tts-adapter-sapinative/src/test/java/org/daisy/pipeline/tts/sapi/impl/OnecoreTest.java
bertfrees: Fixed.

modules/tts/tts-adapters/tts-adapter-sapinative/src/test/java/ignore
bertfrees: Fixed.

modules/tts/tts-adapters/tts-adapter-sapinative/.gitignore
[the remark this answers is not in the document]
NPavie: for the log i don't recall exactly but i think it was the error/crash log of the jvm when a memory leak occured on the native side (haven't got one in while since the updates on the memory cleaning in the native code)

For the apt folders, its a visual studio code extension by red hat when working on java project, that adds those folder (i think those are mainly indexes)

modules/tts/tts-adapters/tts-adapter-sapinative/sapinative.sln
bertfrees: @NPavie do you want to update the README?

bertfrees: @NPavie This PR is almost ready to merge, only this small thing left.

bertfrees: Great. Thanks!

bertfrees:
Merged in https://github.com/daisy/pipeline-modules/commit/9f0cfc5ceb4475d60520d3c1d6b6148fb80ff920.

PR #667, opened 2023-04-03
Various fixes for SAPI tts adapter

NPavie:
This PR intends to fix the following issues detected in latest tests of the adapter in production : 

- In presence of namespaces in the SSML, SAPI could refuse to parse SSML with marks in it : 
  - fixed in https://github.com/daisy/pipeline/commit/a5d07b3aef4dafec55fa0cb2d0db56a24e43f6d0 by removing namespaces when transforming SSML with xsl
- SAPI and OneCore can raise empty bookmarks events during the speech process 
  - fixed in https://github.com/daisy/pipeline/commit/8510fdbd3b3ab904030bde1ff8b1b9a41d1cb414 by filtering marks event that have no associated duration
- The expansion of the internal connections registries of OneCore and SAPI (used for cleanup on closure) could lead to a memory access violation exception
  - fixed in https://github.com/daisy/pipeline/commit/29f10fa339a0fd73a51cf9050ce5b1c6c7685e1d by over allocating the connections registries

Additionnaly added:
- Java IO exception throwing from the native code in commit https://github.com/daisy/pipeline/commit/81baf119dd679fac15b6af3f6d0f2e0e4751b22f
- OneCore voices prioritization over SAPI voices when a voice exists in both API added in commit https://github.com/daisy/pipeline/commit/8510fdbd3b3ab904030bde1ff8b1b9a41d1cb414


modules/tts/tts-adapters/tts-adapter-sapinative/src/main/java/org/daisy/pipeline/tts/sapi/impl/SAPIEngine.java
```diff
@@ -93,33 +88,55 @@ public AudioInputStream speak(String ssml, Voice voice, TTSResource resource, Li
...
+			long[] positions = SAPI.getBookmarkPositions(tr.SAPIConnection);
 			float sampleRate = sapiAudioFormat.getSampleRate();
 			int bytesPerSample = sapiAudioFormat.getSampleSizeInBits() / 8;
-			for (long position : bookmarksPositions) {
-				int offset = (int) ((position * sampleRate * bytesPerSample) / 1000);
-				marks.add(offset);
+			for(int i = 0; i < names.length; ++i){
+				int offset = (int) ((positions[i] * sampleRate * bytesPerSample) / 1000);
+				// NP : it happens that SAPI / OneCore sometimes make empty bookmarks
+				// (i have not found why yet)
```
bertfrees, line 117: @NPavie: are they extra bookmarks that don't correspond to a `mark` element in the SSML? It is important that the [contract of `TTSEngine`](http://daisy.github.io/pipeline/api/org/daisy/pipeline/tts/TTSEngine.SynthesisResult.html#marks) remains satisfied.

NPavie: > are they extra bookmarks that don't correspond to a mark element in the SSML? 

Indeed : on some occasions in my test, i was having registered "bookmark" events with no duration nor name assigned, and that were not linked to a mark in the ssml (all marks were also correctly registered and reported by the connector with their name and duration).

This was causing error in the process where the pipeline was raising an error stating there were more marks than excepted.

bertfrees: OK thanks

bertfrees:
Merged in https://github.com/daisy/pipeline-modules/commit/9f9547c05af9d62eb348aa3e061bf69deb5d44d9

PR #686, opened 2023-06-12
Portage of pipeline 1 dtbook cleaning routines

NPavie:
Reintagrating pipeline 1 dtbook cleaning routines in pipeline 2.

The routines logic is stored in dtbook-utils modules while it is exposed as a pipeline script with a new dtbook-cleaner script modules.


modules/scripts-utils/dtbook-utils/pom.xml
```diff
@@ -64,7 +76,8 @@
   
   <properties>
     <expose-services>
-      org.daisy.pipeline.modules.impl.Module_dtbook_utils
+        org.daisy.pipeline.modules.impl.Module_dtbook_utils,
+        org.daisy.pipeline.modules.dtbook_utils.impl.DTBookCleanerLibrary$Provider
```
bertfrees, line 88: I will move this to org.daisy.pipeline.dtbook.saxon.impl.

modules/scripts-utils/dtbook-utils/src/main/resources/xml/dtbook-utils-library.xpl
```diff
@@ -18,5 +18,7 @@
     <p:import href="validate-dtbook/validate-dtbook.xpl"/>
     <p:import href="dtbook-to-mods-meta.xpl"/>
     <p:import href="dtbook-update-links.xpl"/>
+
+    <p:import href="fix-dtbook/fix-dtbook.xpl"/>
```
bertfrees, line 22: I think I will move the dtbook-cleaner script inside the dtbook-utils module so that

- we don't need to create a new module that we will delete afterwards
- we don't need to expose the new steps to the library yet

For now the dtbook-cleaner script will not be actually exposed as a Pipeline script yet, but I will create a custom release with the script enabled for Save-as-DAISY.

modules/scripts-utils/dtbook-utils/src/main/resources/xml/fix-dtbook/fix-dtbook.xpl
```diff
@@ -0,0 +1,111 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<p:declare-step xmlns:p="http://www.w3.org/ns/xproc" version="1.0"
+                xmlns:px="http://www.daisy.org/ns/pipeline/xproc"
+                xmlns:cx="http://xmlcalabash.com/ns/extensions"
+                xmlns:xs="http://www.w3.org/2001/XMLSchema"
+                type="px:dtbook-fix" name="main">
+    <p:documentation xmlns="http://www.w3.org/1999/xhtml">
+        <h1 px:role="name">Fix DTBook XML</h1>
```
bertfrees, line 8: I know that this happens in other steps too, but these `px:role` attributes only really serve a purpose within script definitions. In a commit that I haven't pushed yet I've removed (most) `px:role` attributes from within the documentation of non-script XProc files.

modules/scripts-utils/dtbook-utils/src/main/resources/xml/fix-dtbook/fix-dtbook.xpl
```diff
@@ -0,0 +1,111 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
bertfrees, line 1: Can you include some documentation to explain that this code came from Pipeline 1 originally?

modules/scripts-utils/dtbook-utils/src/main/resources/xml/fix-dtbook/fix-dtbook.xpl
bertfrees: So this is better:

```xml
<h1>Fix DTBook XML</h1>
<p>Apply fixing routines on a given dtbook xml.</p>
```

modules/scripts-utils/dtbook-utils/src/main/java/org/daisy/pipeline/modules/dtbook_utils/impl/DTBookCleanerLibrary.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    public String getDefaultLocale(){
+        return Locale.getDefault().toString().replace('_', '-');
+    }
+
+    public String getDate(){
+        SimpleDateFormat sdf = new SimpleDateFormat("yyyy-MM-dd");
+        return sdf.format(new Date());
+    }
```
bertfrees, line 30: These can be changed to static methods, so that you don't need the `Library:new()` in library.xsl.

modules/scripts-utils/dtbook-utils/pom.xml
```diff
@@ -23,6 +23,14 @@
       <scope>provided</scope>
     </dependency>
 
+    <!-- Required for dtbook-doctyping to work with correct serialized result -->
+    <dependency>
+      <groupId>org.daisy.libs</groupId>
+      <artifactId>com.xmlcalabash</artifactId>
+      <version>1.1.20-p12-98</version>
```
bertfrees, line 30: Version should be updated in bom/pom.xml.

modules/scripts-utils/dtbook-utils/src/main/resources/xml/dtbook-utils-library.xpl
```diff
@@ -20,5 +20,6 @@
     <p:import href="dtbook-update-links.xpl"/>
 
     <p:import href="fix-dtbook/fix-dtbook.xpl"/>
+    <p:import href="fix-dtbook/doctyping.xpl"/>
```
bertfrees, line 23: We already had functionality in Pipeline to set the doctype (with `px:fileset-store`). We're including the step it in this release but we're going to clean it up afterwards.

modules/scripts/dtbook-cleaner/src/main/resources/xml/dtbook-cleaner.script.xpl
```diff
@@ -0,0 +1,164 @@
...
+                exclude-inline-prefixes="#all">
+
+    <p:documentation xmlns="http://www.w3.org/1999/xhtml">
+        <h1 px:role="name">DTBook Cleaner</h1>
+        <p px:role="desc">Cleaning DTBook documents</p>
+        <div px:role="author maintainer">
+            <p px:role="name">Nicolas Pavie</p>
+            <a px:role="contact" href="mailto:pavie.nicolas@gmail.com">pavie.nicolas@gmail.com</a>
+            <p px:role="organization">DAISY Consortium</p>
+        </div>
```
bertfrees, line 19: I will change this to:

```xml
<address>
    Authors:
    <dl px:role="author">
        <dt>Name:</dt>
        <dd px:role="name">Nicolas Pavie</dd>
        <dt>E-mail:</dt>
        <dd><a px:role="contact" href="mailto:pavie.nicolas@gmail.com">pavie.nicolas@gmail.com</a></dd>
        <dt>Organisation:</dt>
        <dd px:role="organization">DAISY Consortium</dd>
    </dl>
</address>
```

modules/scripts/dtbook-cleaner/src/main/resources/xml/dtbook-cleaner.script.xpl
```diff
@@ -0,0 +1,164 @@
...
+                <px:dtbook-doctyping/>
+            </p:when>
+            <p:otherwise>
+                <p:identity />
+            </p:otherwise>
+        </p:choose>
+        <!-- Store on disk -->
+        <p:store>
+            <p:with-option name="href" select="concat(resolve-uri($output-dir),$output-name)"/>
+        </p:store>
```
bertfrees, line 161: We'll replace this code later with `px:fileset-store`.

modules/scripts/dtbook-cleaner/src/main/resources/xml/dtbook-cleaner.script.xpl
```diff
@@ -0,0 +1,164 @@
...
+    <p:option name="output-dir" required="true" px:output="result" px:type="anyDirURI">
+        <p:documentation xmlns="http://www.w3.org/1999/xhtml">
+            <h2 px:role="name">DTBook</h2>
+            <p px:role="desc">The directory that will contain the resulting DTBook.</p>
+        </p:documentation>
+    </p:option>
+
+
+    <!-- Options from dtbook-fix script -->
+    <p:option name="repair" required="false" select="false()" px:type="boolean">
```
bertfrees, line 66: Script option are always of type string. So the `select` attribute must become: `select="'false'"`. The `px:type="boolean"` is for telling the user interface what are the allowed values.

Note that this also means that in the `px:dtbook-fix` call below, in order to pass the options as booleans, the `p:with-option` must become: `<p:with-option name="repair" select="$repair='true'"/>`

modules/scripts/dtbook-cleaner/src/main/resources/META-INF/catalog.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<catalog xmlns="urn:oasis:names:tc:entity:xmlns:xml:catalog" xmlns:px="http://www.daisy.org/ns/pipeline">
+  <uri name="http://www.daisy.org/pipeline/modules/dtbook-cleaner/dtbook-cleaner.script.xpl"
```
bertfrees, line 3: I'm going to drop the `name`. (It is not required for a script.)

modules/scripts/dtbook-cleaner/pom.xml
```diff
@@ -0,0 +1,71 @@
...
+    <dependency>
+      <groupId>org.daisy.pipeline.modules</groupId>
+      <artifactId>nlp-omnilang-lexer</artifactId>
+      <scope>test</scope>
+    </dependency>
+  </dependencies>
+
+  <properties>
+    <expose-services>
+      org.daisy.pipeline.modules.impl.Module_dtbook_cleaner,
```
bertfrees, line 66: I'm going to drop the `Module` implementation. (The Java class won't be generated if I drop the `name` of the script in catalog.xml.)

bertfrees (changes requested):
@NPavie Very nice job! Thank you.

Shall I wait for that more substantial test you were talking about?

NPavie:
@bertfrees sorry just saw your comment.

I don't think i'll have time to add the tests before the next release.
If you want to merge the PR for the upcoming release, i can do a new PR later to include them (but i need some time to prepare them).

modules/scripts-utils/dtbook-utils/pom.xml
bertfrees: Fixed

modules/scripts-utils/dtbook-utils/src/main/resources/xml/dtbook-utils-library.xpl
bertfrees: Removed from library.

modules/scripts/dtbook-cleaner/src/main/resources/xml/dtbook-cleaner.script.xpl
```diff
@@ -0,0 +1,164 @@
...
+            <p px:role="desc">Publisher to be added as dc:Publisher if none is defined in the dtbook</p>
+        </p:documentation>
+    </p:option>
+
+    <p:option name="ApplySentenceDetection" select="false()" px:type="boolean">
+        <p:documentation xmlns="http://www.w3.org/1999/xhtml">
+            <h2 px:role="name">Apply sentences detection</h2>
+            <p px:role="desc">Encapsulate sentences within the document</p>
+        </p:documentation>
+    </p:option>
```
bertfrees, line 121: @NPavie This option does not seem to be used?

modules/scripts/dtbook-cleaner/src/main/resources/xml/dtbook-cleaner.script.xpl
```diff
@@ -0,0 +1,164 @@
...
+    <!-- ***************************************************** -->
+    <p:import href="http://www.daisy.org/pipeline/modules/dtbook-utils/library.xpl">
+        <p:documentation>
+            px:dtbook-fix
+        </p:documentation>
+    </p:import>
+    <p:import href="http://www.daisy.org/pipeline/modules/dtbook-break-detection/library.xpl">
+        <p:documentation>
+            px:dtbook-break-detect
+            px:dtbook-unwrap-words
```
bertfrees, line 33: @NPavie Are these also supposed to be used?

modules/scripts-utils/dtbook-utils/src/main/resources/xml/fix-dtbook/fix-dtbook.xpl
```diff
@@ -0,0 +1,111 @@
...
+    <p:option name="repair" required="false" select="false()">
+        <p:documentation xmlns="http://www.w3.org/1999/xhtml">
+            <h2 px:role="name">Repair the dtbook</h2>
+            <p px:role="desc">Apply repair routines on the dtbook</p>
+        </p:documentation>
+    </p:option>
+    <p:option name="fixCharset" required="false" select="false()">
+        <p:documentation xmlns="http://www.w3.org/1999/xhtml">
+            <h2 px:role="name">Repair - Fix Charset</h2>
+            <p px:role="desc">Fix the document charset (To be implemented)</p>
```
bertfrees, line 39: Is this option not implemented yet? Let's drop it from the script options then (comment it out).

modules/scripts-utils/dtbook-utils/src/main/resources/xml/fix-dtbook/fix-dtbook.xpl
```diff
@@ -0,0 +1,111 @@
...
+            <h2 px:role="name">Tidy - Simplify headings layout</h2>
+            <p px:role="desc">TBD</p>
+        </p:documentation>
+    </p:option>
+    <p:option name="externalizeWhitespace" required="false" select="false()">
+        <p:documentation xmlns="http://www.w3.org/1999/xhtml">
+            <h2 px:role="name">Tidy - Externalize whitespaces</h2>
+            <p px:role="desc">TBD</p>
+        </p:documentation>
+    </p:option>
```
bertfrees, line 59: Are these also to be implemented?

modules/scripts-utils/dtbook-utils/pom.xml
bertfrees: Done

modules/scripts-utils/dtbook-utils/src/main/java/org/daisy/pipeline/modules/dtbook_utils/impl/DTBookCleanerLibrary.java
bertfrees: Done

modules/scripts-utils/dtbook-utils/src/main/resources/xml/fix-dtbook/fix-dtbook.xpl
bertfrees: Fixed

modules/scripts/dtbook-cleaner/src/main/resources/META-INF/catalog.xml
bertfrees: Done

modules/scripts/dtbook-cleaner/src/main/resources/xml/dtbook-cleaner.script.xpl
bertfrees: Done

modules/scripts/dtbook-cleaner/src/main/resources/xml/dtbook-cleaner.script.xpl
bertfrees: Fixed

modules/scripts-utils/dtbook-utils/src/test/xprocspec/test_fix.xprocspec
```diff
@@ -0,0 +1,61 @@
...
+            <x:document type="port" port="result"/>
+        </x:context>
+        <x:expect label="the dtbook 2005-3 should be as expected"
+                  type="custom" href="http://www.daisy.org/xprocspec/custom-assertion-steps.xpl" step="x:compare-except-ids">
+            <x:document type="inline">
+                <dtbook xmlns="http://www.daisy.org/z3986/2005/dtbook/" version="2005-3" xml:lang="fr-FR">
+                    <head>
+                        <meta name="dc:Publisher" content="DAISY Consortium"/>
+                        <meta name="dc:Language" content="fr-FR"/>
+                        <meta name="dtb:uid" content="AUTO-UID-d990e3"/>
```
bertfrees, line 43: This is brittle. We can perhaps change it later with `http://www.daisy.org/pipeline/modules/common-utils/generate-id.xsl`.

bertfrees:
I've squashed your 4 commits, added a cleanup commit, and rebased onto the latest master.

modules/scripts/dtbook-cleaner/src/main/resources/xml/dtbook-cleaner.script.xpl
NPavie: Sorry i had not seen your messages : 
this was to be used for sentence detection in post-processing of dtbook for the addin (readded in last commit)

modules/scripts-utils/dtbook-utils/src/main/resources/xml/fix-dtbook/fix-dtbook.xpl
bertfrees: I've commented out the "TBD" descriptions for now.

bertfrees:
@NPavie Reminder that this PR needs some tests before it can be merged.

NPavie:
@bertfrees i finished a last round in tests on daisy #718 , i have a small update to make on SaveAsDAISY and i'll resume the creation of tests for this PR.

NPavie:
Hi @bertfrees, i added a small set of tests for the cleaning routines and a bugfix in an xsl that i detected while doing the tests.

modules/scripts-utils/dtbook-utils/src/main/resources/xml/fix-dtbook/fix-dtbook.xpl
bertfrees: @NPavie it would be nice to have a comment here that the module was ported from Pipeline 1, and with a mentioning of both you and the original authors.

modules/scripts-utils/dtbook-utils/src/main/resources/xml/fix-dtbook/narrator.xpl
```diff
@@ -0,0 +1,47 @@
...
+    </p:output>
+
+    <p:option name="publisher" required="false" select="''" >
+        <p:documentation xmlns="http://www.w3.org/1999/xhtml">
+            <h2 px:role="name">Publisher</h2>
+            <p px:role="desc">Publisher to be added as dc:Publisher in the dtbook</p>
+        </p:documentation>
+    </p:option>
+
+    <p:xslt px:message="narrator-metadata">
```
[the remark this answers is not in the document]
bertfrees: @NPavie Do these messages need to show up in the job progress log?

modules/scripts-utils/dtbook-utils/src/main/resources/xml/fix-dtbook/fix-dtbook.xpl
bertfrees: @NPavie Is this option going to be implemented or should we drop it?

bertfrees:
@NPavie Thanks for the fixes! Looking much better now. The only issues left are the following:

- One test seems to be failing
- The CLI is having trouble processing the option descriptions that you added. First, the problem was that the description didn't come through at all because the content of the `role='desc'` element is expected to be plain text containing valid markdown. So I changed that in commits 6c56e55 and be30b2d. Then, it turned out that the CLI does not play nice with HTML blocks, so I tried to fix that in commit 3ab8977. But still the output isn't very nice, I guess because the markdown handling of the CLI is so basic. Now my question is: does the description of these options need to be so extensive anyway? Does the end user need to know such details, or is it more aimed at the developer? In case of the former, my suggestion is to move the details to a separate file and link to it, similar to how it's done e.g. in the [dtbook-to-zedai docs](http://daisy.github.io/pipeline/modules/dtbook-to-zedai/doc/). In case of the latter, my suggestion is to move it to the `pxi:dtbook-fix` step instead. (Note that the documentation of regular steps is HTML, not Markdown.)
- The references to Pipeline 1, e.g. "Prepare dtbook for Pipeline 1 narrator": we should try to get rid of these.

NPavie:
@bert i fixed the failing test (forgot to change a title)

I also removed the mention to pipeline 1 narrator in the current documentation, to specify that it's more of a cleanup of the dtbook for audio-synthesis.

IMO, the documentation might be of use for users who would want to know more precisely what the script is doing while cleaning.
A separate doc folder/markdown file, with links to it in the script inner documentation, makes sense.
If its ok with you, i'll move the detailed script documentation to a `doc/index.md` following the dtbook-to-zedai model you mentionned.

Just in case i forget (but I might make a separate issue for it), another test is failing on windows for this module : 
```
Failed tests:
  test_load
    if the base uri starts with file:///
      * FAILURE: this results in a xml:base with only one slash after file:
```
On my machine, the resulting fileset as `xml:base="file:/F:/"` (F: is the disk my terminal is currently positionned on when running the tests)
As the "/" path does not really exists on Windows, it seems like the produced result is correct (but on windows case, the result itself will depends on dev system configuration and from which disk the test is run)
I don't know if you can disable tests on a system basis with XProcSpec  (like excluding a system for specific tests)

bertfrees:
@NPavie Thanks! Yes, moving the detailed script documentation to a doc/index.md is perfect.

Regarding the other failing test: yes, please create a new issue. That's something we want to fix but not in the upcoming release. You can not disable an XProcSpec scenarion based on which platform the tests are running on. If you want to do that you need to put it in a separate XProcSpec file and then mess with src/test/java/XProcSpecTest.java (override the `runXProcSpec` method of `AbstractXSpecAndXProcSpecTest`). But we should avoid that if possible. It's probably better to change the test. For now you could add a `pending` attribute.

modules/scripts-utils/dtbook-utils/src/main/resources/xml/fix-dtbook/fix-dtbook.xpl
bertfrees: @NPavie I don't see Pipeline 1 and the original authors mentioned anywhere yet.

NPavie: (sorry i just see now the comment you made in the review tab)

Noted, will do

modules/scripts-utils/dtbook-utils/src/main/resources/xml/fix-dtbook/narrator.xpl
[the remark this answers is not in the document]
NPavie: (sorry i just see now the comment you made in the review tab)

Noted, will check that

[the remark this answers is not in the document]
NPavie: Ok checked this, indeed the message does not need to be here anymore, i'll delete them.
(i think i added them for debugging at first)

modules/scripts-utils/dtbook-utils/src/main/resources/xml/fix-dtbook/fix-dtbook.xpl
NPavie: I got back the list of authors that have participated to the dtbook fix script.
I will add a comment stating this script is a portage of pipeline 1 se_tpb_dtbookFix transformer followed by the list of authors in order of participation.

NPavie: should be good in 59df9d0bd3f5fa95cfde4dbc632daaa5af352eed

modules/scripts-utils/dtbook-utils/src/main/resources/xml/fix-dtbook/narrator.xpl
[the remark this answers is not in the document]
NPavie: @bertfrees should be good in 4c90bce9cbc334269111c20d77cf7a8c7445176c
(also removed those messages in the other xpls)

modules/scripts-utils/dtbook-utils/src/main/resources/xml/fix-dtbook/fix-dtbook.xpl
NPavie: I'll also add a similar mention with list of authors in other pipeline 1 ported scripts

bertfrees: Thanks!

modules/scripts-utils/dtbook-utils/src/main/resources/xml/fix-dtbook/narrator.xpl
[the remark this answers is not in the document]
bertfrees: Thanks!

bertfrees:
Merged in [ed88343](https://github.com/daisy/pipeline-modules/commit/ed88343f4d09ebfe07e2d7e92a3dae54ec4d8988)

PR #693, opened 2023-07-18
[modules] dtbook-to-rtf: add missing px:type

NPavie:
Adding boolean type to options of dtbook-to-rtf.xpl script to correct the form rendering in UI


bertfrees:
Thank you! Merged in https://github.com/daisy/pipeline-modules/commit/d8a4ec95154fefd4f5b01b6245bb4535bdafec3e.

PR #718, opened 2023-09-19
TTS Adapter Onecore/SAPI : simplification, system voice priority and multithreading bypass

NPavie:
This PR includes a simplification of the native code side, a system wise selected voices prioritization, and a bypass of multithreading for the speak methods of the connector : 
Multithreaded calls to Onecore/SAPI synthesizer triggers a FastFail error on the native side that we are not able to catch and handle. 

The following changes have been made in the attempt to keep the multithreading active and are included in the PR :
 - Global variables are removed to concentrate declaration as much as possible within functions to try optimizing thread data isolation, 
 - Java Voices are now directly returned by native code adapters instead of being reconstructed from various native method calls that were using global variables
 - Speak methods are now directly returning a new NativeSynthesisResult java object holding both the audio data bytes buffer and the marks names and positions   
 
 Finally, until an investigation on how to re-enable multithreading capabilities is done (possibly with coroutines in winrt), the Onecore speak method calls are now set as synchronized on Windows 11 to avoid concurrent calls to OneCore API that are triggering the fastfail exception.


modules/tts/tts-adapters/tts-adapter-sapinative/src/main/java/org/daisy/pipeline/tts/onecore/Onecore.java
```diff
@@ -10,7 +10,7 @@ public class Onecore {
 
 	public static native Voice[] getVoices() throws IOException;
 
-	public static native NativeSynthesisResult speak(String voiceVendor, String voiceName, String text) throws IOException;
+	public static native synchronized NativeSynthesisResult speak(String voiceVendor, String voiceName, String text) throws IOException;
```
bertfrees, line 13: Could the `synchronized` possibly be conditional on whether we're running on Windows 11 or not?

Also, if the methods are synchronized, perhaps it would be better if the `reservedThreadNum` method of `SAPIEngine` would return 1?

NPavie: > Also, if the methods are synchronized, perhaps it would be better if the reservedThreadNum method of SAPIEngine would return 1?

Ha yes it could be an option too (i went to the fastest way to do it for tests, i did not looked into TTSengine superclass).

i'll try reducing the number of thread that way by checking for windows 11.
(I just need to check how to find out windows version)

bertfrees: Note that the `reservedThreadNum` method can not a replacement of the `synchronized`. Even if the number of threads is set to 1, other TTS threads could still use the SAPI engine in case of a voice fallback. Also, multiple jobs could be running at the same time.

NPavie: So i did a small check : the system property are reporting the same "os.version" between windows 11 and windows 10 (which is 10.0)
But the os.name property returns "Windows 11" in my test, so i'll use this to check for windows version.

NPavie: Code has been updated to use a test on os.name property : 
For Onecore speak call, if os version ends with "11", the calls to the Onecore native method are synchonized on the SAPIEngine class object.

Just in case i tested with the legacy SAPI engine on windows 11 and the error does not occur with it in multithreading context on Windows 11 (the legacy engine is still present on the Windows 11 dev VM they provide) so it's only an issue within the latest version of their Onecore fork of their engine.

NPavie:
Branch is now rebased on latest version of master, everything
I also noticed and fixed a small bug in my detection of voice age on the native side.

modules/tts/tts-adapters/tts-adapter-sapinative/src/main/java/org/daisy/pipeline/tts/sapi/impl/SAPIEngine.java
```diff
@@ -17,15 +17,10 @@
 import net.sf.saxon.s9api.XdmNode;
 
 import org.daisy.common.file.URLs;
-import org.daisy.pipeline.tts.onecore.Onecore;
-import org.daisy.pipeline.tts.onecore.OnecoreResult;
-import org.daisy.pipeline.tts.onecore.SAPI;
-import org.daisy.pipeline.tts.onecore.SAPIResult;
+import org.daisy.pipeline.tts.onecore.*;
```
bertfrees, line 20: I prefer imports of classes or static methods, not whole packages.

modules/tts/tts-adapters/tts-adapter-sapinative/src/main/java/org/daisy/pipeline/tts/onecore/Onecore.java
bertfrees: OK, nice!

What about the `reservedThreadNum` idea? Could still be a good complementary solution I think.

modules/tts/tts-adapters/tts-adapter-sapinative/src/main/java/org/daisy/pipeline/tts/sapi/impl/SAPIEngine.java
NPavie: Intellij tends to rewrote import this way, i'll update that

modules/tts/tts-adapters/tts-adapter-sapinative/src/main/java/org/daisy/pipeline/tts/onecore/Onecore.java
NPavie: I can add it, but i don't know how it will change the current behavior
(or maybe it will reduce thread locking on the jvm side so that could reduce latency on job execution for this case ?)

NPavie: Note that this would also limit multithreading for SAPI legacy

modules/tts/tts-adapters/tts-adapter-sapinative/src/test/java/org/daisy/pipeline/tts/sapi/impl/OnecoreTest.java
```diff
@@ -23,58 +26,23 @@ public static void load() throws SynthesisException {
...
-		Onecore.closeConnection(connection);
-	}
 
 	/**
 	 * Test if voices can be recovered with the TTS engine
 	 * @throws Throwable
 	 */
 	@Test
 	public void getVoiceInfo() throws Throwable {
+		System.out.println(System.getProperty("os.version"));
```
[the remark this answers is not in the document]
NPavie: done

modules/tts/tts-adapters/tts-adapter-sapinative/src/main/java/org/daisy/pipeline/tts/sapi/impl/SAPIEngine.java
NPavie: done

modules/tts/tts-adapters/tts-adapter-sapinative/src/main/java/org/daisy/pipeline/tts/onecore/Onecore.java
NPavie: `reservedThreadNum` is now overriden to return 1 on Windows 11 and 0 otherwise

modules/tts/tts-adapters/tts-adapter-sapinative/src/main/java/org/daisy/pipeline/tts/sapi/impl/SAPIEngine.java
bertfrees: Intellij does this? Hmm I see.

modules/tts/tts-adapters/tts-adapter-sapinative/src/main/java/org/daisy/pipeline/tts/onecore/Onecore.java
bertfrees: > I can add it, but i don't know how it will change the current behavior

It might not change much in practice, especially when SAPI is the only TTS engine that is used. It's just that if you know in advance that only one sentence can be spoken at a time, it makes sense to do it all in one thread, in order to have as little overhead as possible. Probably not noticeable, but still.

> Note that this would also limit multithreading for SAPI legacy

Ah yes, I didn't think of that. Perhaps a little comment regarding this is appropriate.

bertfrees:
@NPavie I cleaned up the branch a bit:
- rebased
- squashed two commits
- updated commit messages
- made every commit compile

My only remaining question is about the second commit: what is the effect/goal of putting the default voice first in the list?

modules/tts/tts-adapters/tts-adapter-sapinative/src/main/java/org/daisy/pipeline/tts/onecore/Onecore.java
bertfrees: Added a comment

bertfrees (approved):

NPavie:
> @NPavie I cleaned up the branch a bit:
> 
> * rebased
> * squashed two commits
> * updated commit messages
> * made every commit compile
> 
> My only remaining question is about the second commit: what is the effect/goal of putting the default voice first in the list?

ok good !

In a discussion a while ago i think you mentionned the order if insertion of voices in the list could influence its priority.
For people without configuration, the idea is to first use the system-wide selected voice if possible.

But it's a bit complicated to manage the adapter main voice here as there are two settings competing with each other on windows : one for the default SAPI voice and one for the default Onecore voice.
For now i prioritize SAPI as some users might have third party voices installed for it (like acapela voices for windows).

bertfrees:
> In a discussion a while ago i think you mentionned the order if insertion of voices in the list could influence its priority.

Hmm, I don't know if that's the case. (And don't recall mentioning that.) It could be the case, but then it is probably an implementation detail. We would need to document the feature and add some unit tests for it.

> For now i prioritize SAPI as some users might have third party voices installed for it (like acapela voices for windows).

Do you mention that somewhere in the code?

NPavie:
> Do you mention that somewhere in the code?

No its not mentionned, but the SAPI Voices are loaded first so i assume they are prioritized.

Given the problem we are encountering with onecore multithreading, i'm wondering if we should not keep all voices available (included the SAPI versions of the one that are also declared in OneCore, we are currently discarding the sapi version of a voice if a onecore version exists).
@bertfrees what do you think ?

bertfrees:
I mean the fact that you prioritize SAPI (load it first) because users might have third party voices installed for it: It would be good to mention that so that we remember why it is done.

Regarding keeping all voices: what would that mean in practice? The reason you removed the duplicates was to hide the implementation details from the user. If we keep the duplicates, the user will have the power to force SAPI or OneCore, but at the cost of more complexity. You can argue that Pipeline should automatically choose the underlying implementation that works best.

NPavie:
I added a comment in the code to explain a bit why sapi voices are prioritized.

If nothing else, i think we can merge

bertfrees:
Very nice!

bertfrees:
> Before we can merge, we still need to document and test that the order of the getAvailableVoices list has an influence on voice selection.

@NPavie There is a problem. I tested this and it appears the order has no influence on voice selection. So your "System wide selected voices prioritization" commit does not have the desired effect. I will see what I can do.

bertfrees:
OK, fixed it. See my latest commit.

NPavie:
Last minor change:
I expanded the locking timeout in native code after a test in production of the adapter.
I stumbled upon a case where the release of the mutex lock took one or two seconds more than the 3 seconds i defined in the code.
(The conversion did not crash, but one sentence was synthesize with a fallback onecore voice immediately after the error arise in the log.)

Just in case, i extended the timeout test to 10 seconds.

I did more tests and it did not happended again so far after this extension of the timeout.

bertfrees:
OK good! Let me know if it's ready. I will merge soon.

NPavie:
I think it is finally ready, I don't have crahs anymore on windows 11, performances are a bit lower than before but the impact is contained as much as possible, so I would say go for it :)

Issue #77, opened 2014-04-13
Xpath function registry


ghost:
_From [capitan....@gmail.com](https://code.google.com/u/118277064934808789355/) on July 28, 2011 17:55:45_

Create a xpath function registry adding them to calabash.

_Original issue: http://code.google.com/p/daisy-pipeline/issues/detail?id=77_

PR #807, opened 2025-01-14
fix set-doctype: close the input stream reader before replacing the file

NPavie:
Quick fix for an error in the SetDoctypeProvider class on Windows system (as reported in daisy/pipeline-modules#77) :
A Files.move operation was done to replace a file that was still opened through an InputStreamReader object and locked by the OS.

(tested in production, no error reported after the fix)


bertfrees:
Merged in [2382b8dd10](https://github.com/daisy/pipeline-modules/commit/2382b8dd10)

PR #824, opened 2025-05-20
Text-to-speech adapter for Amazon Polly

NPavie:
This PR integrates the text-to-speech adapter for Amazon Polly engines into the DAISY Pipeline 2 modules tree.

Original code was authored by @mmartida and provided by [ONCE](https://www.once.es/) to the DAISY Pipeline team.
The adapter was then updated for the latest version of Daisy Pipeline 2 and refactored to match similar adapters structure (mainly azure and google TTS adapters).

The adapter has also been updated to allow the use of generative, long-form, or neural voices if available for the user's AWS account.

The following properties are used by the adapter :
- `org.daisy.pipeline.tts.aws.accesskey` (mandatory) : Access key to connect to Amazon Web Service.
- `org.daisy.pipeline.tts.aws.secretkey` (mandatory) : Secret key to connect to Amazon Web Service.
- `org.daisy.pipeline.tts.aws.region` (mandatory) : Region associated with your Amazon Web Service account.
- `org.daisy.pipeline.tts.aws.priority` (defaults to `15`) : priority of usage within the pipeline if other text engine are available

The adapter relies on AWS java Polly client library. All dependencies required for its usage are added to the `modules/bom` project.

In case it would be preferable to release the adapter on maven before including it's first release version, this PR :
- Does not register the adapter itself in the `modules/bom`
- Does not include the aforementioned dependencies and the new adapter in the current assembly project

(@bertfrees I'll leave you the squashing and rephrase for the pipeline history depending on how you want it merged in)

### DIFF
--- a/modules/tts/tts-adapters/tts-adapter-aws/src/main/java/org/daisy/pipeline/tts/aws/impl/AWSPollyTTSEngine.java
+++ b/modules/tts/tts-adapters/tts-adapter-aws/src/main/java/org/daisy/pipeline/tts/aws/impl/AWSPollyTTSEngine.java
@@ -1,0 +1,183 @@
+package org.daisy.pipeline.tts.aws.impl;
+
+import java.io.ByteArrayInputStream;
+import java.io.IOException;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.Locale;
+import java.util.Map;
+import static java.util.stream.Collectors.toList;
+
+import javax.sound.sampled.AudioFormat;
+
+import com.google.common.io.ByteStreams;
+
+import net.sf.saxon.s9api.SaxonApiException;
+import net.sf.saxon.s9api.XdmNode;
+
+import org.daisy.common.file.URLs;
+import org.daisy.pipeline.tts.TTSEngine;
+import org.daisy.pipeline.tts.TTSRegistry;
+import org.daisy.pipeline.tts.TTSService.SynthesisException;
+import org.daisy.pipeline.tts.Voice;
+import org.daisy.pipeline.tts.VoiceInfo.Gender;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import software.amazon.awssdk.auth.credentials.AwsCredentials;
+import software.amazon.awssdk.core.exception.AbortedException;
+import software.amazon.awssdk.regions.Region;
+import software.amazon.awssdk.services.polly.PollyClient;
+import software.amazon.awssdk.services.polly.model.DescribeVoicesResponse;
+import software.amazon.awssdk.services.polly.model.SynthesizeSpeechRequest;
+import software.amazon.awssdk.services.polly.model.Engine;
+import software.amazon.awssdk.services.polly.model.TextType;
+import software.amazon.awssdk.services.polly.model.VoiceId;
+import software.amazon.awssdk.services.polly.model.OutputFormat;
+
+/**
+ * TTS adapter for Amazon Polly
+ * @author mmartida - original author
+ * @author Nicolas Pavie - updates and refactoring for DAISY Pipeline 2, version 1.15+
+ */
+public class AWSPollyTTSEngine extends TTSEngine {
+
+	private final static URL SSML_TRANSFORMER = URLs.getResourceFromJAR("/transform-ssml.xsl", AWSPollyTTSEngine.class);
+
+	private final static Logger LOGGER = LoggerFactory.getLogger(AWSPollyTTSEngine.class);
+
+	private static final AudioFormat AF_PCM_POLLY = new AudioFormat(AudioFormat.Encoding.PCM_SIGNED, 16000f, 16, 1, 2, 16000f, false);
+
+	/**
+	 * Client object provided by AWS SDK to interact with the polly engines
+	 */
+	private final PollyClient client;
+
+	/**
+	 * Map of AWS voices available to the user for each engine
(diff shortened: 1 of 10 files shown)


modules/tts/tts-adapters/tts-adapter-aws/src/main/java/org/daisy/pipeline/tts/aws/impl/AWSPollyTTSService.java
```diff
@@ -0,0 +1,91 @@
...
+				return Optional.of(Integer.valueOf(str));
+			} catch (NumberFormatException e) {
+				throw new SynthesisException(str + " is not a valid a value for property " + prop.getName());
+			}
+		}
+		return Optional.empty();
+	}
+
+	@Override
+	public String getName() {
```
bertfrees, line 83: In the commit message you mentioned this:

> Change name of engine from "polly" to "aws" for mapping in UI project

What exactly is this for?

bertfrees:
@NPavie I'm finally having a look at your PR. I have rebased the branch onto master and did some minor tweaks to the code formatting, and commit messages. I also removed some commented out code. These all seemed to stem from the original code (which is still available on the [polly](https://github.com/daisy/pipeline/compare/polly) branch).

Incidentally, I have also rebased the original branch to [polly-rebased](https://github.com/daisy/pipeline/compare/polly-rebased), and am checking whether there could be some useful stuff that isn't in this PR.

First thing I notice (apart from all the changes in that branch not related to the AWS adapter) is that there were a lot of tests. I haven't looked at them yet. Were they not useful you think?

I will have some more remarks on specific pieces of code.

modules/tts/tts-adapters/tts-adapter-aws/src/main/java/org/daisy/pipeline/tts/aws/impl/AWSPollyTTSEngine.java
```diff
@@ -0,0 +1,207 @@
...
+					}
+				}
+			}
+			for (Engine engine : engines) {
+				if(awsVoicesDescription.containsKey(engine)) {
+					voices.addAll(
+						awsVoicesDescription.get(engine).voices()
+							.stream()
+							.map(i -> new Voice("aws", i.idAsString() + " (" + engine.toString() + ")",
+							                    new Locale(i.languageName()),
```
bertfrees, line 131: In other places locales are parsed like this:

```java
(new Locale.Builder()).setLanguageTag(info.getLocale().replace("_", "-")).build()
```

In the original code there was some language normalisation going on.

- `arb` was replaced by `ar` with the same meaning.
- `cmn-CN` was replaced by `zh_CN`

(Originally this transformation was done in `VoiceInfo.tagToLocale)(`, not in `TTSEngine.getAvailableVoices()`, but the result should be the same.)

I wonder whether this normalisation is still needed. Note that we have the `org.daisy.pipeline.common.NormalizeLang` class (in common-utils) which could _in theory_ be used for it.

modules/tts/tts-adapters/tts-adapter-aws/src/main/java/org/daisy/pipeline/tts/aws/impl/AWSPollyTTSService.java
NPavie: If i recall, the UI project uses the engine "name" to extract and organize the settings in the engine settings view (i think it filter settings by the presence of this name in settings key)

modules/tts/tts-adapters/tts-adapter-aws/src/main/java/org/daisy/pipeline/tts/aws/impl/AWSPollyTTSEngine.java
```diff
@@ -0,0 +1,207 @@
...
+	}
+
+	/**
+	 * Get the list of available voices
+	 *
+	 * @throws SynthesisException
+	 * @throws InterruptedException
+	 */
+	@Override
+	public Collection<Voice> getAvailableVoices() throws SynthesisException, InterruptedException {
```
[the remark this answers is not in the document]
NPavie: We are providing more info now : `getAvailableVoices` now also include the underlying polly engine in the name of the voice (because some voices can be available in several engines)

modules/tts/tts-adapters/tts-adapter-aws/src/main/java/org/daisy/pipeline/tts/aws/impl/AWSPollyTTSService.java
bertfrees: Is it something that the engine's `nicename` would be more suitable for? This should be the only name that the user sees in the user interface.

bertfrees:
Merged in 6dbdb2f5d9.

bertfrees:
I was able to create an account a run the tests succesfully.

It was not clear from the README where I can find the region identifier, so I just took the same as you (eu-west-3).

One thing I noticed was that there are debug messages on standard out/err. This is because of the logback.xml file in src/test/resources. @NPavie Did you do this on purpose?

NPavie:
Hi @bertfrees 

For the region identifier, if you open the "IAM Identity center" of your account (not the "IAM" service, those are 2 different service) you should see the region identifier in the parameters summary on the left.
I also noticed in my case that the region identifier is also displayed in the URL of the console as `region=eu-west-3` url parameter.


For the logback report, I might have left it for debug. You probably can remove it if it is too verbose.
(Also if i recall the amazon library is itself very verbose)

